PR #18, opened 2016-01-21
Fixed a number of compiler warnings

hogthrob:
- Missing returns and potential uninitialized use of variables by adding proper return statements, declarations or defaults
- Changed some code to use break and if instead of goto to prevent  uninitialized use of variable n1
- Added some volatile keywords and casts to make compiler (and user) happy.


df8oe:
You start working great :) What's your callsign?

vy 73
Andreas

PR #19, opened 2016-01-21
Revert "Fixed a number of compiler warnings"

df8oe:
Reverts df8oe/mchf-github#18
tuning is broken with this work so I reverted. You have now big gaps in each band - Please take a look at it!


df8oe:
Do you build binary and test them "live" on a mcHF? There are many issues which cause compiler warnings but when you remove or fix them binary does no longer work correctly! So I must know if you program without testing in mcHF or if you test binaries...

vy 73
Andreas

PR #20, opened 2016-01-21
Fixes for compiler warnings, nothing serious

hogthrob:


df8oe:
I cannot merge this. Since your last work (where si570 code was changed) tuning is broken. There are gaps in the guaruanteed working frquencies which where only present after your fixes. I reverted last commit due to this big issue but new pull does have this issue, too.

Sorry - there is a rework needed in si570 code!

vy 73
Andreas

hogthrob:
I fixed the regression in the si570 code, fixed missing {} and a misunderstanding of the original code. Now works (at least I am able to sweep through the bands without problems).

PR #21, opened 2016-01-21
Devel df8 oe 0.0.219.26.x

hogthrob:
This time it should be okay, the regression is fixed and it seems to work properly.


df8oe:
OK - now it seems to work. I have merged it and will test it next hours on my mcHF. Mine is running every day because of I am self-employed and it is standing near my working place :)

vy 73
Andreas

PR #22, opened 2016-01-22
Array-out-of-bounds

hogthrob:
Sorry for the messy pull request for just 2 lines of change, I still have to get hold of how to operate github correctly

### DIFF
--- a/mchf-eclipse/drivers/audio/audio_driver.c
+++ b/mchf-eclipse/drivers/audio/audio_driver.c
@@ -10,7 +10,6 @@
 **  Last Modified:                                                                 **
 **  Licence:		For radio amateurs experimentation, non-commercial use only!   **
 ************************************************************************************/
-
 // Common
 #include "mchf_board.h"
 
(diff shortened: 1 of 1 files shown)


df8oe:
Yes working with git is something one must learn and if you use console there are MANY otions you can use. It is a mighty tool - but tricky, too :)

vy 73
Andreas

df8oe:
BTW:
Do you understand the working of the variables test_a etc.? I cannot believe this are compiler issues or STM library faults... May be it is related to "out of bounds" errors, too. This would be a very important work to eliminate the variables, and may be compiler optimizations will work then, too. At the moment any compiler optimization whic is set leads to not-working builds - but *it gets builds*!!

vy 73
Andreas

hogthrob:
BTW, I can build and run (!) with -03 if I turn off optimization for few files. 

From my experience the compiler gets the code right even with strong optimization stages more than 99,99%. In most cases it is an exposure of the human factor (wrong code) if optimization breaks "working" software. As for the test_xxx -> these should go away since they are a strong sign of bugs somewhere, most likely some of these are connected to the issues I fixed. 

The fixes I provide are all based on inspection of compiler warnings. Once these are properly fixed, we have a good chance to identify the reamaining issues which break the optimization more easily. 
I expect most of these issue in improper use of direct memory interaction with devices (for instance I had to turn off optimization for mchf_i2c2.c to get it to run)

PR #23, opened 2016-01-22
More small fixes and restructurings

hogthrob:
Please review changes in menu system. As far as I can see and tested, it should good (I am not sure if it is ever called with item numbers above 48, this will now properly calculate a menu location which it didn't before).

### DIFF
--- a/mchf-eclipse/drivers/ui/ui_menu.c
+++ b/mchf-eclipse/drivers/ui/ui_menu.c
@@ -52,6 +52,9 @@
 // CW generation
 #include "cw_gen.h"
 //
+#include "mchf_hw_i2c2.h"
+
+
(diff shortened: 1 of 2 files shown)


df8oe:
Yes I will do a short review and testing. I have commented out test_x variables at your code and binary works :) So I think you have fixed this by removing out-of-boundary-errors.

I can give you an advice where problem in i2c2.c can probably be found:
I must use sequential writing mode of serial EEPROM because of bytewise write is *very very* slow. So I must grab some RAM where data which should be written to is stored. Unfortunatley it is necessary that this RAM is accessible in different functions. I have tested malloc() but it seems to be not working correct in STM library. Please take a look first at this - this may be the issue that causes optimization run into darkness...

vy 73
Andreas

PR #24, opened 2016-01-22
Fixes a few warnings

hogthrob:
, mostly to make code more robust in case of calling with
invalid arguments.
Fixes a logic expression bug in UiDriverChangeBand


df8oe:
Very nice work. Now I got only 1 warning related to "__packed" - but I think this is a gcc issue of definition twice (I compiled in IDE Eclipse for comparing to some stored "warning files"). All other warnings are gone and my mcHF is running without curious "test_x" variables since over 2 hours. I will push this version to "testing" this weekend because it is a milestone now.

vy 73
Andreas

PR #27, opened 2016-01-23
Fixed issue in copying values to/from eeprom (array-out-of-bounds)

hogthrob:
Please review this carefully. The fixes was made without having an EEPROM available but should be good if indeed only MAX_VAR_ADDRS [currently 383] configuration values are supposed to carry meaningful data and should be stored. It also now contains the bootload issue fix. Now the test_*  vars can go to where they belong...

### DIFF
--- a/mchf-eclipse/Makefile
+++ b/mchf-eclipse/Makefile
@@ -19,7 +19,7 @@ GCC_LIB_VERSION = 4.8
 
 # compilation options
 CFLAGS  = -mcpu=cortex-m4 -mfloat-abi=hard -mfpu=fpv4-sp-d16 -mthumb -ffunction-sections -O0 \
- -DARM_MATH_CM4 -DSTM32F407VG -DSTM32F4XX -DUSE_STDPERIPH_DRIVER
+ -DARM_MATH_CM4 -DSTM32F407VG -DSTM32F4XX -DUSE_STDPERIPH_DRIVER -Wall $(EXTRACFLAGS)
 LDFLAGS := -mcpu=cortex-m4 -mfloat-abi=hard -mfpu=fpv4-sp-d16 -mthumb
 LIBS := -Llibs -larm_cortexM4lf_math -L$(PREFIX)/arm-none-eabi/lib/armv7e-m/fpu \
  -L$(PREFIX)/lib/gcc/arm-none-eabi/$(GCC_LIB_VERSION)/armv7e-m/fpu -lm -lg -lc -lgcc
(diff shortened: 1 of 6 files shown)


hogthrob:
I added also my Makefile changes, should have no impact on existing users on Linux. Enables easier working with OpenSTM32 (which to I intend to switch from CoIDE because of better git at least compared to my old CoIDE)

df8oe:
I have tested mcHF under real conditions with outcommented test_ variables since yesterday and I did not get any problems. So I am sure issues are gone now. Now it will be very interesting to investigate why compiler optimizations break function. Many issues regarding this are gona already - it starts, and most functions work already - but not all...

I must revert one format parameter you switched from "u" to "ld" because it has broken function (display frequency at menu #213) - but it was very easy to fix and you have done a very good job!

vy 73
Andreas

hogthrob:
To play with the optimization I recommend to place  #pragma GCC optimize("O0") inside the files. So you can control the place where optimization is (not) being done. See https://gcc.gnu.org/onlinedocs/gcc/Function-Specific-Option-Pragmas.html .

df8oe:
Nice option - I didn't know that :)

Will you play around on this or shall I do? It's not necessary that both of us are working on that. I would start it - I am interested in optimization :)

vy 73
Andreas

hogthrob:
I have some other area I'd like to look into (SPI display speed) so you can play with it. I did that and noticed that all c files in the audio/ part and the mchf_i2c*.c cannot be optimized with O3 right now. Some of the problems I noticed might be gone now with my bootloader issue fix, haven't tried that again.

df8oe:
OK I start to play with it. SPI LCD enabling would be very interesting for getting more free GPIOs...

vy 73
Andreas

Astralix:
I see a lot of changes there. One makes me somewhat nervous :)
The Windows optimizations probably break some Linux compatibility in the Makefile. The rule is to provide the path right into the /bin/arm-none- way. So normally the Makefile should use
A default:
CROSS_COMPILE ?= /opt/gcc-arm-none-eabi-5_2-2015q4/bin/arm-none-eabi-
( ?= sets this value only if you did not set it in your global exports)

CC := $(CROSS_COMPILE)gcc
LD := $(CROSS_COMPILE)gcc
...

I'd like to change that and I'd like to silence the Makefile a bit more, so you see the errors and warnings a lot better

hogthrob:
Feel free to find a better solution in the Makefile for the Windows part. This is indeed just a hack from my side. Should not have any impact on Linux. 

The  other change is -Wall which I recommend to keep. Most if not all of the issues  I fixed I found by working my way through the list of warnings. Good software does not produce warnings when compiled (well even if this means to disable a "stupid" warning local occasionally). The gcc does not issue warnings for fun.  BTW, software which compiles without warning is not necessarily good software, I am aware of that.

PR #29, opened 2016-01-25
Slight cleanup of the (SPI) LCD code

hogthrob:
, replaced all single byte transfers of data with calls to the respective functions.
The Spi*Fast functions are not in use in this commit (but used in my testing).

Moved the delay in a section with disabled optimization at the end.
This may or may not solve the parallel display issue regarding optimization
with O3. Due to lack of parallel display I cannot test this.

No relevant changes for parallel mode.

Cosmetic change to ui_menu.c

### DIFF
--- a/mchf-eclipse/drivers/ui/lcd/ui_lcd_hy28.c
+++ b/mchf-eclipse/drivers/ui/lcd/ui_lcd_hy28.c
@@ -48,20 +48,7 @@ extern __IO   SpectrumDisplay      sd;
 
 extern __IO OscillatorState os;		// oscillator values - including Si570 startup frequency, displayed on splash screen
 
-//*----------------------------------------------------------------------------
-//* Function Name       : UiLcdHy28_Delay
-//* Object              :
-//* Input Parameters    :
-//* Output Parameters   :
-//* Functions called    :
-//*----------------------------------------------------------------------------
-void UiLcdHy28_Delay(ulong delay)
-{
-   ulong    i,k;
-
-    for ( k = 0 ;(k < delay );k++ )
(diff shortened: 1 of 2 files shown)


Astralix:
My display should be connected in parallel so I'll test tonight. (On the weekend I forgot all my JTAGs at the company)
I report back.

df8oe:
OK - tested. Is running in SPI faster than before and no effects in parallel mode. Optimizations are not finished yet - if it is enabled for files under /usb and/or under /cmsis_lib firmware crashes as before. I have to investigate what's the problem and will do that the next hours/days. I choose O3 for testing with much regressive code so that faults are seen immediately (I hope so). I know it is not optimum setting for later. But at this stage I first must investigate where problems are located and then put O0 to specific files which must not be optimized at all and remove all other pragmas then.

Bug when I optimize ui_lcd_hy28.c behaviour is very strange.... Screen is completely ok and working, and the first minutes ALL is working correctly. When you leave mcHF alone for ~20 minutes, no button is working any more - no touchscreen (TP_IRQ is interpreted as "button"). But all encoders work and rx and tx do work, too.... You only can disconnect power and then same procedure occurs after a few minutes of running. With O0 all is working fine.

vy 73
Andreas

df8oe:
Works succeeded faster than I thought first...

Only two files cannot be optimized:
/drivers/ui/lcd/ ui_lcd_hy28.c (issue pse have a look at the last post)
/cmsis_lib/source/stm32f4xx_i2c.c (issue no sound and crunched spectrum/waterfall data)

I set these two files for O0 and removed #pragma from all other files. Optimization now is set to O2 in Makefile.

vy 73
Andreas

PR #30, opened 2016-01-25
Hopefully improved the UI performance:

hogthrob:
Changed BulkWrite to do indeed a BulkWrite of a number of pixels from an
array (used in WaterfallDisplay)
And replaced almost all SPI transfer with the more optimized *Fast.

Please test carefully. In SPI mode I can use O3 w/ no problems.
Cannot test parallel with O3. My hunch is that the SRAM timing is now
too fast and should be reconfigured to be slow since there are less
instructions inbetween calls and the display controller may not be
able to keep up with the speed of the MCU unless given some break.
Should be fixable by changing the SRAM interface timing.

Added changes to the menu system asking for testing and comments, see commit message for  6690f1c28f72d4dc2c504480673de8837c7abe69


df8oe:
I will test carefully. I do have different mcHFs here running in parallel and SPI mode so I can test everything. First impress in both modes is good. SPI is working much faster and reactions of buttons and encoders is improved strongly. In parallel mode WF and spectrum are running in a first quick test, too. I will do tests next morning.

vy 73
Andreas

hogthrob:
Please note that git added more changes in a different area (Menu System) to this pull request.
Please comment on the changes to ui_menu.c as asked for in the commit message. Thanks!

Astralix:
Hi! @hogthrob , could you check if the display doesn't accept 16bits transfer, cause AFAIK the stm32 is able to use this kind of transfer. I'm currently setting up semihosting debug and then check the optimization issue and IIC.

hogthrob:
@Astralix: I am not sure what you mean with 16 bit transfers? Do you mean, 16 bit transfers in parallel mode? Or in SPI mode? To be clear: hardwarewise I cannot check anything other than SPI with HY28A, since I happen to own a HY28A which is fixed to SPI. My HY28B is somewhere lost in space.

df8oe:
As I understood parallel mode exists in 8bit and 16bit mode, where mcHF is wired as 16bit mode. SPI Mode always is 8bit. You can set device ID bit by setting/unsetting IM0.
New release runs well with -O2 both iin parallel as in SPI. Speed is improved significantly in SPI mode.

@hogthrob: Is there some comment lost? which commit do you mean?

vy 73
Andreas

hogthrob:
@df8oe : Well SPI is technically not tied to a specific data length. The display controller is somewhat agnostic to the bit length in an individual transfer as long as the data is finally of the correct length. Since all transfers the controller expects are multiple of 8bits, this is used. We cannot use 16bits, since some transfers are of 24 bits length. Anyway: I am happy that it works well even in parallel mode.

As far as I see it, with the current software architecture speed of SPI (or parallel for that matter) cannot be improved much more, since we most likely run the display now close to its limits and have to wait for the data transfers to be completed and cannot do much more work in between (expect for serving the interrupts). 
Only area where there could be some real gain is the display update for the waterfall / spectrum but it may be some major task to convert it to a different approach using DMA for instance. Not sure if it is worth it right now.

df8oe:
I think at the moment lcd speed is ok as it is. Cleanig up code e.g. for menu will get code much smaller. Next may be to have  a look if we can switch over to a more effective task scheduling. At the moment only one or two interrupts are used as I saw but I did not investigate completety.

My next tasks would be trying to implement freedv (the problem there is a different sample rate than in mchf), wspr (passive and active) (the problem here is we must get samples @200Hz for 120seconds == ~100KB of data and PSK31 (I think only much work in keyboard integration, and ui implementation) also creating memory (50...100 of them) for storing stations to get "quick access".

vy 73
Andreas

hogthrob:
I think, at the moment there is still a need for cleaning up the foundations so that one in future can concentrate on the new stuff without having to worry that the base is too buggy and interfers with the improvements. 

What I intend to look at after the cleanup is the CAT interface which is right now completely broken and my test code has very strange lookups (although TX/RX switching using DTR works on Linux hosts, not on Windows hosts). That would fit nicely with your work for digital modes which I am interested in as well.

df8oe:
I agree. Cleanup is neccessary. CAT interface was broken somewhat in the past. It never is remarked as "working" since over 1 year. Clint never touched the code I think it is done by Chris in the r&d-time of mcHF. A working CAT will be nice, too. I cant imagine more uses for that: decoding via external SDR-program and tune mchf via CAT...

vy 73
Andreas

PR #32, opened 2016-01-27
Fixed a regression in the UiMenu introduced in last commit

hogthrob:
Ser EEPROM Settings were displayed even if no EEPROM avail.

### DIFF
--- a/mchf-eclipse/drivers/ui/ui_menu.c
+++ b/mchf-eclipse/drivers/ui/ui_menu.c
@@ -307,7 +307,7 @@ void __attribute__ ((noinline)) UiDriverMenuMapColors(uint32_t color ,char* opti
 
 
 
-static char* base_screens[10][MENUSIZE] = { {
+static char* base_screens[10][MENUSIZE] = { { //1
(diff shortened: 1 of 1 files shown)


df8oe:
The only informations about working of firmware I have is the pdf from Clint located in the useful_manuals folder. There are code blocks I would have done completely different - think we must try&error a little bit at some functions...

vy 73
Andreas

df8oe:
Now menu text EEPROM #197 and #198 are not displayed even if serial EEPROM is installed...

vy 73
Andreas

hogthrob:
Thanks, I have been looking at it for very long, but testing is difficult without a serial eeprom (in fact, I don't know why me eeprom is not showing up, but this is a different story).

df8oe:
Not a problem. What type of EEPROM do you use? Only a few devices are supported and probably you have to do a hardware mod if you use 24LC1025. 24LC1026 works out-of-the-box and AT24CM02 (2MBit type) should too but I don't have one for testing. 24LC512 works out of the box, too. That are the only EEPROMS that are supported.

vy 73
Andreas

PR #35, opened 2016-01-28
Fixed regression in menu items 207, 209 210

hogthrob:
See pull request title, 209 and 210 was simple copy'n'paste issue. 207 is more difficult.
Maybe my solution to use "4" for off (anything less then 5 will be treated as "off" but "4" will be stored) is not acceptable since it is a slight change (with no real impact) but still a change.


df8oe:
OK - I have seen it already. I fixed all and selected "0" as OFF - pse pull again!

df8oe:
all fixed - please pull again

vy 73
Andreas

PR #36, opened 2016-01-28
Refactoring of EEPROM Save Code

hogthrob:
to use functions instead of code clon…es all over.

see void UiDriverSaveEepromValuesPowerDown(void)

Gets a little boring after the first 50+ refactorings, so I will call it a day for today...
But saved a few bytes (in fact I think more than a kByte...) so in total we
can expect to loose some 4kBytes.

And some minor layout fixes for AFG, RIT, RFG using right aligned text

### DIFF
--- a/mchf-eclipse/drivers/ui/ui_driver.c
+++ b/mchf-eclipse/drivers/ui/ui_driver.c
@@ -5911,7 +5911,7 @@ static void UiDriverChangeAfGain(uchar enabled)
 		UiLcdHy28_PrintText((POS_AG_IND_X + 1), (POS_AG_IND_Y + 1),"AFG",Grey1,Grey,0);
 
 	sprintf(temp,"%02d",ts.audio_gain);
-	UiLcdHy28_PrintText    ((POS_AG_IND_X + 38),(POS_AG_IND_Y + 1), temp,color,Black,0);
+	UiLcdHy28_PrintTextRight    ((POS_AG_IND_X + 55),(POS_AG_IND_Y + 1), temp,color,Black,0);
 }
 
 //*----------------------------------------------------------------------------
@@ -6177,7 +6177,7 @@ void UiDriverChangeRfGain(uchar enabled)
 
 		sprintf(temp," %02d",(int)ts.fm_sql_threshold);
 	}
-		UiLcdHy28_PrintText    ((POS_RF_IND_X + 32),(POS_RF_IND_Y + 1), temp,color,Black,0);
+		UiLcdHy28_PrintTextRight    ((POS_RF_IND_X + 55),(POS_RF_IND_Y + 1), temp,color,Black,0);
 
 }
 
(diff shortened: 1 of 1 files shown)


df8oe:
I don't know if this was already present in the past or it is a new issue - but pse take a look at this:
- if I toggle through DSP modes color of #310...#315 should change correspondently
- it is not possible to adjust #313 and #314 - regardless of anything
- it is possible to change #313...#315 even if DSP is off (should only adjustable & white when corresponding DSP is on)
- you can decrease #315 to 0 what results in firmware crash

I hope that is the only resting issue in menu system :)

vy 73
Andreas

hogthrob:
I will look into this, but currently have no access to a mcHF just laptop and compiler until tomorrow

df8oe:
ok - can wait. I want to push .26 to stable this weekend and when this will be fixed, it will be great. But I think only a few OMs have changed these settings for testing...

vy 73
Andreas

BTW:
Do you have fixed your serial EEPROM issue?

hogthrob:
No, haven't look into it as it does not bug me too much. I have a 24LC1025 (now with the proper 3 to 8 wire). Maybe  it was not good that I forgot to wire it in the beginning. 
Anyway, would have to change the code and use a debugger to find out what is going wrong, but this is too much for me :-). 

BTW, I did look at the code for a few moments. I am not sure what goes wrong here. Changing the settings is not guarded by a check of ts.dsp_active. Looking briefly the at the original code, it seems that this was never guarded by a dsp.active check. 

Honestly, the configuration parameter management is not nearly perfect. Due to the high amount of similar code used, it was very easy to get things not right.

I am also not convinced that the "two" level defaults are intended at all. The current code has 3 different place where defaults / ranges are checked/enforced (Load from EEPROM, Save to EEPROM, UI Menu). It is very likely to not get aligned handling of defaults since they are defined in these 3 places independently. 

Once I recover from the my current refactorings I might change that to use a proper daa structure to have that consistently defined in a single place instead of spread around in different places (or someelse does this, I don't mind). This adds to good user experience but not at all to functionality so most developers will not be interested to much in this I guess. And it takes some time...

df8oe:
OK - checks are "forgotten" by Clint. But I remembered I could adjust #314 and #315 (hope I am not wrong). They are fixed now - unchangeable.

The existence of the three different "default" checks leads to inconsistences actually. If you reset mchf back to defaults use F1-F3-F5 at powering on and later have a look what happens if you use "DEFLT" button you will see many changes which should not be there. Unless it will reduce code length and readability significantly. I will start looking into digital modes - if you want to do this rework, there are no dupe works...

vy 73
Andreas

hogthrob:
I found the issue.

hogthrob:
@df8oe I will probably go over the LoadEEPROM function in the next days but I am not sure if I will go the easy route or the hard one. Anyway, this will reduce code by about 3-4 kBytes (estimation based on the savings from Save). The alignment itself will not reduce code much more, since the 3x150 calls will still be there. 
Since I would also like to have some fun, I will as already mentioned look into getting the CAT interface working properly. I already have found a CDC demo project based on  more or less the same STM32 sample code used in mcHF. The demo does work on the STM32F4Discovery board and I try to learn from it why the CDC code on the mcHF stops working after a few successful communications. Once that is working stable, the mcHF development community has to discuss how to best implement a CAT interface.  Don't know which is the best place to start this discussion. Yahoo, german forum, github issue.

df8oe:
Looking at some STM32 code which works on discovery board was my way to write bootloader. So I think you will get the goal walking this way, too. About discussing:

Yahoo group should be bigger and international - only "international" is correct. In our german group are about 120 very active builders and same number passive readers. In the past I saw in Yahoo group there is not a big movement to walk forward - they want to "finish" their mchf thats all. In our german group many OMs are testing and improving actively. I think we must not do a decision. You start discussion in Yahoo group and I start it in our german group. We can gather the results and then we will have a quite solid basement.

vy 73
Andreas

PR #38, opened 2016-01-29
Fixed a regression where the menu did not work properly (313 and 314).

hogthrob:
Was a copy'n'paste issue. Wrong variable was changed.

To check if similar issues are around: the sprintf at the end must use the same variable name which has been used by UiDriverMenuItemChange*Int* . 
The EnableOnOff/DisableOnOff etc. do the sprintf internally, so it will always print the variable being modified. But even this may be the wrong one but this is harder to check. You will have to look into older versions to identify issues here. Normally there is a good correlation between the CONFIG_... label and the variable name.


df8oe:
I must be absolutely blind. This was my first thought about that, too. I looked at the code parts many times and I didn't see it - like it's code made by me.... Normally you do not see your own errors but shortly the errors of others - this is an example it is OFTEN but ALWAYS same ;)

vy 73
Andreas

PR #42, opened 2016-01-31
Replaced the existing USB files with similar ones in order to get a working virtual com port 

hogthrob:
The old code was somehow broken, it stopped working after a few bytes CDC traffic (around 62 to 77).
Was not able to find the issue, replace it with working code from STM32 libs.
There may be issues with host mode (was not enabled, not tested),
audio (also not enabled).
All in all there is no loss in existing, working functionality. Using this
code will enable development of proper CAT remote control for the mcHF.

Currently there is no useful code there. Only an echo when using the virtual
comm port is provided, which allows minimal testing.

### DIFF
--- a/mchf-eclipse/Makefile
+++ b/mchf-eclipse/Makefile
@@ -70,7 +70,6 @@ usb/otg/src/usb_core.c \
 usb/otg/src/usb_hcd_int.c \
 usb/otg/src/usb_dcd.c \
 usb/otg/src/usb_dcd_int.c \
-usb/otg/src/usb_otg.c \
 usb/otg/src/usb_hcd.c \
 usb/usbd/core/src/usbd_core.c \
 usb/usbd/core/src/usbd_ioreq.c \
--- a/mchf-eclipse/drivers/cat/cat_driver.c
+++ b/mchf-eclipse/drivers/cat/cat_driver.c
@@ -20,7 +20,7 @@
 #include "usbd_usr.h"
 #include "usb_conf.h"
 #include "usbd_desc.h"
-
+#include "usbd_cdc_vcp.h"
 #include "cat_driver.h"
 
 __ALIGN_BEGIN USB_OTG_CORE_HANDLE    USB_OTG_dev __ALIGN_END ;
@@ -41,10 +41,6 @@ extern __IO CatDriver kd;
 	//EXTI_ClearITPendingBit(EXTI_Line18);
 //}
 
-void OTG_FS_IRQHandler(void)
-{
-	USBD_OTG_ISR_Handler (&USB_OTG_dev);
-}
 
 void cat_driver_init(void)
 {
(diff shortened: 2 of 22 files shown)


df8oe:
Maybe here is something incomplete - i cannot build firmware with this files. I use Makefile - but I don't think that is the cause

usb/otg/src/usb_otg.c: In function 'USB_OTG_GetCurrentState':
usb/otg/src/usb_otg.c:402:14: error: 'USB_OTG_CORE_HANDLE' has no member named 'otg'
   return pdev->otg.OTG_State;


vy 73
Andreas

hogthrob:
Sorry, you pointed to the right thing: remove the usb/otg/src/usb_otg.c from the Makefiles. It is not used and it breaks the build.

df8oe:
I tried that and it doesnt fix the build-fail. Removing the usb_otg.c causes error:

drivers/cat/usb/usbd_bsp.o: In function `OTG_FS_IRQHandler':
usbd_bsp.c:(.text.OTG_FS_IRQHandler+0x0): multiple definition of `OTG_FS_IRQHandler'
drivers/cat/cat_driver.o:cat_driver.c:(.text.OTG_FS_IRQHandler+0x0): first defined here

Beneath these error are many warnings related __packed__ and some related to wrong function arguments (not urgent at the moment). But building using Makefile must work - I think you have quicker way to solution because you can take a look in your build log what is different to Makefile...

vy 73
Andreas

-----

Your idea about working on CAT is excellent - I see it same way. Protocol of FT817 is well documented and easy and includes everything you need. I support this and I do not think further discussion is needed...

vy 73
Andreas

df8oe:
ISSUE:
Loading of bandspecific settings (mode, freq_a, freq_b) does not work. VFO-A is loaded correct, VFO-B always shows "0000000" in freuency. At a short review this is not related to saving but to loading. I think there must be three different functions - or some more parameters because arrays f a and b are named xxx_a and xxx_b. You have well thought about the functions and maybe you have a tricky solution :) Some minor bugs are fixed - all c&p errors or easy to track condition errors. Not so much if I regard the amount of code which you have reworked!

vy 73
Andreas

hogthrob:
Well, regarding the USB commit: I checked in a working cat_driver.c (with the removed interrupt handler). I know about the warnings, these are the same we had before my clean-up. I will go over them, this time I'll take it more slowly to verify that my changes do not break the usb support.

The VFO  fix is also in this pull request. Should be easy to isolate and to apply on other branches.

hogthrob:
The isse should not exist in master, as it is using older code,  which does not have the refactored vfo code.

hogthrob:
Please wait a moment, I saw an issue.

df8oe:
I start ca. 19:00 UTC.

For discussing internally you can contact me via email om@df8oe.de. Sometimes it is easier and does not fill GitHub ;)

vy 73
Andreas

PR #44, opened 2016-02-01
Added simple FT817 emulation. Just enable CAT mode in config menu.

hogthrob:
Tested with fldigi/wsjt-x/HamRadioDeluxe.
A number of commands are NOPs.
Had to refactor a function into two in ui_driver to allow proper setting of
new demodulation.

### DIFF
--- a/mchf-eclipse/drivers/cat/cat_driver.c
+++ b/mchf-eclipse/drivers/cat/cat_driver.c
@@ -1,15 +1,15 @@
 /************************************************************************************
-**                                                                                 **
-**                               mcHF QRP Transceiver                              **
-**                             K Atanassov - M0NKA 2014                            **
-**                                                                                 **
-**---------------------------------------------------------------------------------**
-**                                                                                 **
-**  File name:                                                                     **
-**  Description:                                                                   **
-**  Last Modified:                                                                 **
-**  Licence:		For radio amateurs experimentation, non-commercial use only!   **
-************************************************************************************/
+ **                                                                                 **
+ **                               mcHF QRP Transceiver                              **
+ **                             K Atanassov - M0NKA 2014                            **
+ **                                                                                 **
+ **---------------------------------------------------------------------------------**
+ **                                                                                 **
+ **  File name:                                                                     **
+ **  Description:                                                                   **
+ **  Last Modified:                                                                 **
+ **  Licence:		For radio amateurs experimentation, non-commercial use only!   **
+ ************************************************************************************/
 
 // Common
 #include "mchf_board.h"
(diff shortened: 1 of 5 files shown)


hogthrob:
Regarding the issue: After changing the frequency switching through bands is not working properly, it shows the changed frequency (with the correct band in the position of the original band. After reboot the "default" frequency is shown.

df8oe:
I got CAT running - but there is one big issue... If CAT once was enabled, you can no longer use any button on mcHF. Encoders are working, but you cannot leave menu. This remains when disabling CAT again. Only solution is removing power supply.

Maybe speed of ui interface is decreasing very strong - but to test this buttons must be usable first :)

y 73
Andreas

hogthrob:
That is strange. I have fully working buttons with CAT enabled. I was transmitting using WSJT-X for some time and still all buttons work. Only difference I can think of is the display but I cannot see how this could lead to the button not working scenario.

df8oe:
That *is* strange. I just have tested on second mcHF and there buttons work! The one where buttons work is running SPI LCD, the other is running parallel. Please stay tuned for a few minutes I will test furthermore.

vy 73
Andreas

df8oe:
OK - its located. If LCD is in SPI mode all is working fine. But if LCD is parallel buttons function is broken. It is 100% reproduceable.

vy 73
Andreas

df8oe:
Very nice. Linux is recognizing as serial port out-of-the-box and working as inspected. After a short test using SPI LCD I do not see great differences in speed if CAT is enabled or not.

vy 73

Andreas

hogthrob:
PA9 and PA10 have been "used" by the USB code. I turned these off.

df8oe:
ok. intensive testing can start now!

hogthrob:
There is serious issue with the current implementation: It requires all communication received in 5 byte increments. The original protocol uses some sort of timeout to reset its buffer, even if only 4 bytes have been transmitted. My implementation does have this (but needs it). So if you send 1 extra byte, the cat interface will never recover from this. Since there is no checksum or anything else, the code has no way to tell if the 5 bytes it reads are a proper command or not. Is not much of an issue with USB and well written clients but removing cables in mid of transfer and such things may lead to a  blocking.

I pretty sure, we should enable some kind of global clock in the mcHF and use it.

df8oe:
I am not sure I want to call it "issue". But it is something we can (and should) do a workaround...

Many devices stop working when you pull off cables when transfers are in progress...

hogthrob:
IMPORTANT: Regarding CAT mode on Windows: It is important to install the STM32 VCP driver on  the composite device and not on the 2 COM ports showing up in the Device Manager as unknown devices. If you see 2 COM Ports in the Device Manager, it will not work. Uninstall both drivers and install (manually) on the composite device shown under the USB hub (Show by connection helps to identify the "right" composite device).

df8oe:
Thanks - I have noticed this already. I put STM driver in our German project group in discussion forum and we will start testing.

PR #45, opened 2016-02-02
Fixed compiler warnings

hogthrob:


df8oe:
It seems the only one which has survived is the damned __packed__ warning :)

I have feedback from three users who have tested CAT. No issue reported.

hogthrob:
Yes, but I had a fix for that as well in the "old" usb code. We'll get that back again. I don't see this message in CoIDE, only when I run the OpenSTM32 makefile compiler (I can't get this to debug the compiled binary with STLINK / GDB / OpenOCD, otherwise I would switch ASAP).

PR #46, opened 2016-02-02
Disabled __packed redefined warning

hogthrob:


df8oe:
There is a CAT feature missing - but I cannot reproduce because I am not an active CW'er...

There is an option (e.g. in UCX-log) that you can send CW text to transceiver and it interprets and transmits in CW. As I understand feature wish, there is a "VOX" option missing. Sorry -I can not sescribe better way I have git screenshots from the user who asks for the feature. If you see a possibility to add this feature I need your emailadress to send you the screenshots.

Andreas

PR #48, opened 2016-02-03
CAT Mode settings is now being stored in EEPROM.

hogthrob:
CAT Mode starts when booted if enabled in config.

Maybe the choice of the number is not clever, but that can be changed.

And issue with not storing (in fact not reading back) negative values from flash/eeprom. This issue is not in the master as it uses older code, which did not have this regression.


df8oe:
Please make a pull first there was an issue in mch_board.h (missing #define)

Issue #49, opened 2016-02-03
Feature: Make LINEIN Channel Selection an configuration option.


hogthrob:
MCHF uses by default the left channel data for line in. Like in my case, the right channel would be the right channel. I would suggest to implement this as additional option in Menu 060 Mic/Line Select to be now MIC / LINE-L / LINE-R with MIC being the default. This would be 100% backward compatible but would solve my problem.

df8oe:
Thats absolutely ok. Feature does not afect anything and maybe helps others, too.

vy 73
Andreas

hogthrob:
Implementation is done, only issue I could imagine is the slight change in the UI interaction:
M3 long press now toggles through "MIC"-"LIN"-"L-R" since I opted to reuse the existing tx_audio_source parameter. That means you need two long press instead of one to get back to microphone. I don't know if this creates a usability issue. If so, it can be implemented with an additional  parameter which tells which line to use and you have to select the left/right in the config menu. With this switching between MIC and LINEIN and selecting the LINEIN channel would be separate choices. Both ways have its pros and cons.

PR #50, opened 2016-02-04
Added support for issue #49 (Linein Channel Selectable)

hogthrob:
Changes: 060 now has settings MIC - LINE-L - LINE-R.
Stored settings are 100% compatible, i.e. if previously LINE was selected now LINE-L is display which is exactly the same thing. 
Long press on M3 now toggles through the 3 choices instead of two. L-R is displayed if right channel LINEIN is selected.


df8oe:
Not a problem when you must push twice.

vy 73
Andreas

PR #61, opened 2016-02-14
added make output files into .gitignore

jkramarz:
These binary files, compiled using "make", should not be tracked by Git.


df8oe:
Folder firmware_Targe_Flash is completely in gitignore so there is no need to add files within this folder.
---
Changes are only done in devel-branch (except slight functional bugfixes which may be removed in master and testing seperately to devel) - please use devel branch for further investigations.
---
Python bytecodes can be removed of course. But they were part of the of the package I found on the web for building dfu from hex and they are only a few bytes long. For giving complete information related to hex2dfu I decided to leave all here.

vy 73
Andreas

jkramarz:
@df8oe, yes, the firmware_Target_Flash directory is in .gitignore, but it's not used by Makefile - the result files are placed in build root.
Shall I rather encourage Makefile to use it than add result files to .gitignore?

Also, talking about Python bytecode - the problem is just the same: these files are getting recompiled after building bootloader, so displays Git considers them changed. There's no difference for Python if they're not present in repository clone.

I'd like to get rid of untracked files from repository clone left after compiling using bare GNU Make.
These changes may be confusing when looking at repo history.

Should I rebase to devel-df80oe or did you mean the "testing" branch?

df8oe:
Please rebase to devel-DF8OE.

OK - now I understand your acting and will add it - but to devel-DF8OE.

Such changes are implemented in devel branch and step through testing to master - so we have best security for bugs being fixed before thea reach master...

vy 73
Andreas

df8oe:
I have merged your pull request manually to devel-DF8OE

vy 73
Andreas

PR #62, opened 2016-02-14
I2C Refactoring hopefully done right

hogthrob:
Larger than normal commit. Contains more or less completely refactored I2C based on the changes made for the EEPROM code recently.
It should be more stable and uses less code. All I2C access operations are now based on the same low level base functions which try to implement the STM32F4 datasheet information as best as possible. May have still gotten something wrong. Please search for bugs (and fix them :-) )

Also some minor changes elsewhere due to changed return value data size.


df8oe:
I'll do testing now.

vy 73
Andreas

PR #66, opened 2016-02-15
Added support information for running live debugging with Eclipse

db4ple:
GNU Eclipse ARM tool chain.


jkramarz:
Hi @db4ple; may I suggest using already existing `useful_manuals` directory?

db4ple:
Yes, maybe for the images. Don't care too much about the where, information should be integrated into a build and debugging manual eventually and the images can go away. However the more important MCHF.cfg does not belong to the manuals as it is needed. Was looking for a place to store extra information which is not source code somewhere outside the root directory.

PR #70, opened 2016-02-16
First real sound coming out of USB audio.

db4ple:
WORKS. Needs testing.


df8oe:
Very nice! I will place binary in our German discussion group - not in "daily snapshots". Reason is that CAT is lost at this stage and as I notice many OMs use CAT already. But testing will be done - I am sure :)

vy 73
Andreas

PR #76, opened 2016-02-20
AUDIO OUT: Works much, much better now.

db4ple:
Luckily it was bad weather today :-) 
See my last commit comment, there are some issues still but in general it works quite good.
Two new input modes have been added (switchable in Menu #060 and with long press on the button under encoder 3): DIGITAL (DIG) and DIG I/Q (DIQ).
Digital uses the audio from the left channel for modulation according to the set mode (SSB, AM, ... ).
Digital I/Q uses both channels to transmitt IQ data 48khz, 16 bit. The second choice is untested (which program could I use?) but should work (since it essentially just bypasses all (!) the processing).
Only when in DIG or DIG IQ the USB audio out channel is being used. Otherwise only the sound  is sent to PC and  of course CAT FT817 emulation is running all the time once CAT mode is enabled.
Overhead of the USB DIGITAL modes mode is roughly  1,5% (or 15uS each 1ms). So the device should be as responsive as ever.

### DIFF
--- a/mchf-eclipse/drivers/audio/audio_driver.c
+++ b/mchf-eclipse/drivers/audio/audio_driver.c
@@ -1834,10 +1834,20 @@ static void audio_tx_processor(int16_t *src, int16_t *dst, int16_t size)
 	int16_t				*ptr;
 	uint32_t			pindex;
 
+	// If source is digital usb in, pull from USB buffer, discard line or mic audio and
+	// let the normal processing happen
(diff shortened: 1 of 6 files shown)


df8oe:
You can test USB-I/Q functions with SDR# or HDSDR running Windows.HDSDR should be capable of RX and TX, SDR# only RX...

vy 73
Andreas

PR #84, opened 2016-02-23
Various small refactorings.

db4ple:
Use of ColorMap function in most places
Use of audio_xlate_freq() to calculate freq shift in all places
Simplified dial digit display function
Small other things


db4ple:
I just noticed there is an issue in this commit, Frequency Display is not correct. In one of the frequency translations I must have done something seriously wrong. Will not be able to look at it now.

PR #95, opened 2016-02-27
fixed sd variable name double use in global scope

jkramarz:
The name "sd" was used for SoftDds and SpectrumDisplay in the global scope.
Has someone said that structure member doesn't work for some crazy reason? ;-)

This change has not been tested on real hardware; I need two more weeks to gather parts, as I'm out of funds.


db4ple:
:+1:  for you and this finding

PR #105, opened 2016-02-28
Removed inaccessible code or code without result

jkramarz:


df8oe:
Maybe there is a problem related to 700b7e0 but I am not 100% sure. I will review this later.

vy 73
Andreas

PR #106, opened 2016-02-28
Reduced some variables scope and assignments without result

jkramarz:
By the way, is it really better to use unsigned int?
http://stackoverflow.com/questions/4889707/compiler-optimization-on-marking-an-int-unsigned

### DIFF
--- a/mchf-eclipse/drivers/ui/ui_driver.c
+++ b/mchf-eclipse/drivers/ui/ui_driver.c
@@ -6234,7 +6234,6 @@ static void UiDriverReDrawSpectrumDisplay(void)
 		//
 		case 3:
 		{
-			uint32_t 	i;
 			float32_t		filt_factor;
 			//
 			filt_factor = (float)ts.scope_filter;		// use stored filter setting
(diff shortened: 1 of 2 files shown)


db4ple:
Regarding data type choice: I like to use the data type which provides best clarity of the intent and most simple code. For instance unsigned types can be used to avoid writing checks for negative values of parameters, i.e. anything used as indicies into arrays should be unsigned.
The width of the data type is an different thing entirely. In many places I like to have an explicit width data type used. But which is the right one is not always clear to me. So this choice may be a bit random. With an 32bit processor, in most cases performance can benefit from using a 32bit or 16bit data type based on the concrete operation. However, for most place in the mcHF Software this discussion is kind of useless as in almost all cases, it is not the performance killer. Wait for a single byte of SPI transfer takes at 25Mhz clock rate (this is the one used) about 330ns or in CPU Cycles about 56. And we do transfer a lot of bytes to the screen every second. I guess, it is not much better with the parallel mode (which uses a external SRAM interface), it also rans with a much slower speed compared to the normal RAM (I didn't check what exact timing is used, you may try to figure this out).

But this is all guess work or possibly wrong interpretation of the real thing. Only way to find out is to profile the mcHF application properly. And to identify the performance weakspots. Most software does have a very tiny amount of code/architerual block which consumes an unexpected amount of resources. However, once you removed those problems, gaining performance gets very though. 
And then it would be easier to switch to a new hardware platform which simply executes faster. Like the STM32F7. Don't get me wrong, I believe we are far from this point right now. I would like us to concentrate to find the "performance eaters" and deal with them first.

db4ple:
BTW, I am all in for declaring local index counters in the 'for' loop instead defining them  elsewhere.

df8oe:
Please do not declare variables within a for-loop because this implements changing of compiler settings. Declaring inside for-loops throws errors. I will have a look at this pull-request later.Here in DL the sun is shining pretty much and this is quite rarely the last weeks so that I will spend time with my family now :)

df8oe:
OK - let's change compiler option that errors will be gone. I will check this later.

vy 73
Andreas

jkramarz:
@db4ple, talking about STM32F7... I think that using pin-compatible STM32F429VI with Chrom-ART Accelerator would be enough - see TouchGFX demos, it's crazy!

PR #112, opened 2016-02-28
USB Improvements and Ui Driver Improvements

db4ple:
Edit: All tested now, and works for me.
These changes are simple, but  I could not test them right now. If no time to check, please leave them.
But in case reports of USB audio issues on Windows, please consider at least the USB change for inclusion.

Due to some more small changes, the pull request has grown a bit. 

It will break the other pull request #115 since both touch the Frequency Encoder. (Or vice versa).
@df8oe: I would suggest that you test out #115 and this request independently. I would be able to merge the code of the other pull request into my branch and commit a new pull request with the merged implementation (of course you can do it by yourself, it is not going to be hard).

### DIFF
--- a/mchf-eclipse/drivers/cat/usb/usbd_desc.c
+++ b/mchf-eclipse/drivers/cat/usb/usbd_desc.c
@@ -42,7 +42,7 @@
   * @{
   */ 
 #define USBD_LANGID_STRING              0x409
-#define USBD_MANUFACTURER_STRING        (uint8_t*)"STMicroelectronics"
+#define USBD_MANUFACTURER_STRING        (uint8_t*)"mcHF Open Source Community (based on STMicroelectronics Drivers)"
 
 #define USBD_PRODUCT_HS_STRING          (uint8_t*)"STM32 Virtual ComPort in HS mode"
 #define USBD_SERIALNUMBER_HS_STRING     (uint8_t*)"00000000050B"
@@ -50,7 +50,7 @@
 #ifndef USB_AUDIO_SUPPORT
 	#define USBD_PRODUCT_FS_STRING          (uint8_t*)"STM32 Virtual ComPort in FS Mode"
 #else
-	#define USBD_PRODUCT_FS_STRING          (uint8_t*)"STM32 AUDIO in FS Mode"
+	#define USBD_PRODUCT_FS_STRING          (uint8_t*)"mcHF USB Interface"
 #endif
 #define USBD_SERIALNUMBER_FS_STRING     (uint8_t*)"00000000050C"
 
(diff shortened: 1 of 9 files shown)


df8oe:
This version has a major bug. Frequencyis counted continously without user action and LPF relais trigger at 2Hz.

vy 73
Andreas

db4ple:
Thanks,

as I said, untested and not really suited for inclusion except for the
USB part.

Will look into this.

df8oe:
It's all ok - not a problem. I think with this description you'll fix it very fast. I think it's a "major bug" because it makes mcHF unusable but I think also it only a very small change that makes it work again :)

I am absolutely curious about USB changes and wanted to test them imediately but now "I have to wait " ;) I do not want to cherry-pick only one part or merge it manually. I think problem is in rotary modifications.

vy 73
Andreas

db4ple:
The main change for Linux is already in previous commits (merged
already). This commit should fix a small Windows regression.

73
 Danilo

df8oe:
I know.... A well working Windows AND Linux soultion would be very fine!

vy 73
Andreas

db4ple:
Added fix for Encode Issue (still not tested but it was an obvious mistake). Thanks Andreas, description of issue was perfect. Took only a little thinking to find the issue.

df8oe:
Hello Danilo,

it is not possible to build this. With existing settings I got a

error: 'for' loop initial declarations are only allowed in C99 or C11 mode

If I try compiler option for c11 or c99 mode, it builds, but function is broken. Something that changes when using c11 or c99 mode (which is mandatory for initial declarations) breaks code at other places.

vy 73
Andreas

db4ple:
Hello Andreas,

it is for sure not an issue regarding C11 mode, I strongly believe the 
issue is in my own code. Why? I have been compiling with C11 enabled for 
quite a while (since I switched from CoIDE to Eclipse, all Eclipse user 
compile with C11). It was another small issue in my commit I am sure. 
See my final commit.
Simply try to compile without my commits and C11. Should work perfectly. 
Maybe it works with my last commit as well.

73
  Danilo


Am 29.02.2016 um 14:12 schrieb DF8OE:

df8oe:
I cannot agree.

Existing devel on GitHub builds & works perfectly without C11 or C99 mode. But built with C11 or C99 it builds without errors but function is broken.

For testing use existing version on my GitHub and uild with -std=c11 or -std=c99

Both binaries show horizontal lines instead of waterfall or scope (other issues not tested).

vy 73
Andreas

db4ple:
Strange, as I said I have been compiling with -std=c11 under Eclipse the whole time.  Hm. Don't like this. Needs investigation. Which compiler version do use?

df8oe:
Hello Danilo,

I use Makefile from repo and

gcc-arm-none-eabi                                           4.9.2-10+14~bpo8+1

Thats the newest productive usable version which is available for Debian Jessie using backports.

I never could build project using c11 or c99. I used older compiler versions before - never successful with these options.

vy 73
Andreas

db4ple:
@df8oe : I updated my front comment for this pull request.

db4ple:
Thanks.

df8oe:
I have tested USB audio onmy Linux system and on my Windows XP. Everything works fine using different software (fldigi, wsjtx, qsstv).

Do you want to create a new pull request or shall I take a look at merging manually?

vy 73

Andreas

PR #118, opened 2016-03-01
Added VOLUME CONTROL to supported Audio Interface Features for in and…

db4ple:
… out

signal. Just a DUMMY for testing if it appears in the various operating
systems correctly as a control. Volume does not change, not matter what you set there.

### DIFF
--- a/mchf-eclipse/usb/usbd/class/audio/inc/usbd_audio_core.h
+++ b/mchf-eclipse/usb/usbd/class/audio/inc/usbd_audio_core.h
@@ -111,6 +111,7 @@
 #define AUDIO_STREAMING_INTERFACE_DESC_SIZE           0x07
 
 #define AUDIO_CONTROL_MUTE                            0x0001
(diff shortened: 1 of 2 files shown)


df8oe:
Hello Danilo,

I have built & tested. Notihng changes here. Audio in & out works as before, device is not recognized by any mixer (pulseaudio, alsa) - so no volume control visible.

vy 73
Andreas

db4ple:
OK, but Windows seems to be happy. Well, I'll integrate real volume control for digital i/o in mcHF and this should at least give users the ability to adjust volume manually. And we'll get the other stuff working as well. I am sure.

df8oe:
100% agree. No need for volume controls running Linux. Every application I use (fldigi, wsjtx, qsstv) has its own volume setting and all of these work perfectly. Let us focus Windows users. We must have a look at Win10 - may be there is a problem but I do not have reported it of any other user than Jos. I have started a "positive" lookup - we will see ;)

vy 73
Andreas

PR #119, opened 2016-03-01
USB Volume control and more volume control simplifications/extensions

db4ple:
This a hell of a commit. Had to add different tx_audio_source handling 
this is very much streamlined now. Allowed to add settings for TX gains for
all different tx sources individually. Added rx_usb_gain to control digital
line out (indepent from audio gain and tx gain).
Right now only the ts.rx_usb_gain is connected to the USB volume control.
I.e. in Windows you can now control the input volume using the
controls slider. Not yet a proper mapping of volume value and gain
(linear value used instead of logarithm, but that is a piece of cake).
Because of the changes, please test especially transmitting with all
different audio sources (mic, line in left, right, digital)...

There will be issues here most likely as I tested only very little.


df8oe:
I'll first build firmware using your sources and play around with it. I report if I have tested.

vy 73
Andreas

df8oe:
Everything works fine as before, no probems in audio input assignment, USB level check will continue.

WANTED: Somebody who uses Windows 10 for testing.

vy 73
Andreas

PR #122, opened 2016-03-02
Code Restructuring: Spectrum and Waterfall

db4ple:
Moved all code and almost all constants related to computing and
drawing spectrum views to ui_spectrum.
Did a few optimizations on the Spectrum Scope code, should run much faster
in most cases.
Enabling the optimization for the ui_lcd_hy28.c gave at least the SPI
display a real performance boost.

Please test using the different displays and modes. If any issue arises, we should try to fix it
without having to give up optimization for LCD code.

### DIFF
--- a/mchf-eclipse/drivers/audio/audio_driver.h
+++ b/mchf-eclipse/drivers/audio/audio_driver.h
@@ -16,8 +16,8 @@
 #define __AUDIO_DRIVER_H
 
 #include "arm_math.h"
-#include "ui_lcd_hy28.h"
-
+// #include "ui_lcd_hy28.h"
+#include "ui_spectrum.h"
 // 16 or 24 bits from Codec
 //
 //#define USE_24_BITS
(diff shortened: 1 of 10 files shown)


db4ple:
I am experiencing button lockups right now, unfortunately I was debugging at this time. Switched step size. Then it locked, it is an issue related to the step size button, I can reproduce this.Step Size- does the trick.

df8oe:
I have located problem - but I do not understand!!!

On PC4 pullup resistor is MISSING (!!!) If you press button GPIO is pulled down and stays down. Sometimes voltage floats (without touching button) gently down and when it reaches switching level step size changes by chance. Few seconds later voltage is "L" and the button check is blocked :)

I have checked GPIO_Structures and init routines but all seems to be ok. With older release pullup is working. I do not think you have changed something in init structures so only possibility is that something is written to portc register at a place by accident. I think this is a *very good* hint ;)

vy 73
Andreas

PR #125, opened 2016-03-03
Added hispeed SPI mode for HY28B displays.

db4ple:
For HY28B: Clocks the SPI for bus with 50Mhz instead of 25Mhz as it is necessary
for the older HY28A display. Works perfectly on my display. YMMV.

NEEDS TESTING!

### DIFF
--- a/mchf-eclipse/drivers/ui/lcd/ui_lcd_hy28.c
+++ b/mchf-eclipse/drivers/ui/lcd/ui_lcd_hy28.c
@@ -8,7 +8,7 @@
 **  File name:                                                                     **
 **  Description:                                                                   **
 **  Last Modified:                                                                 **
-**  Licence:      CC BY-NC-SA 3.0                                                **
+**  Licence:      CC BY-NC-SA 3.0                                                  **
 ************************************************************************************/
 
 // Optimization disable for this file
@@ -77,7 +77,7 @@ void UiLcdHy28_BacklightInit(void)
 //* Output Parameters   :
 //* Functions called    :
 //*----------------------------------------------------------------------------
-void UiLcdHy28_SpiInit()
+void UiLcdHy28_SpiInit(bool hispeed)
 {
    GPIO_InitTypeDef GPIO_InitStructure;
    SPI_InitTypeDef  SPI_InitStructure;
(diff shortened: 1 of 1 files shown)


df8oe:
Hello Danilo,

with this version touchscreen is broken (no function). I have three SPI HY28B for testing: two don't work, one does. If I warm up the two not working one works sometimes, but with display errors. mcHF with parallel LCD works (but without touchscreen). I think we are to near or above the specs of LCD controller chip and above specs of touchscreen controller :/

I will investigate more.

vy 73
Andreas

db4ple:
This does not come completely surprisingly. There is no need for a touchscreen controller to be accessed at 50Mhz SPI clock, given the amount of data being transferred to the host.

I think, that we most likely can added dynamic switching of the clock. For the very low number of touchscreen interrupts per second, the slow down due to changing the register back and forth is neglectable. But of course, we could also check, if there is another SPI on the now unused parallel lines and connect the touchscreen controller there. Would not make a difference for the parallel mode and HY28A users, they could continue to use the old one. Even HY28B users can do this, accepting the display lower speed. User with the right mod could be detected and get the touch screen from a different SPI bus. Haven't looked into the datasheet to see if this is possible.

df8oe:
I will remove touchscreen SPI on my testing mcHF and try running LCD in SPI highspeed mode. I have tested already pullups at device and/or MCU side of SPI lines but no difference. LCD is not working stable. But maybe touchscreen controller disturbes signal at high speed - I'll test it.

vy 73
Andreas

db4ple:
Thanks. For parallel mode, we just have to set the SPI bus to low speed again. A simple call to UiLcdHy28_SpiInit(false) will do the trick. Switching the speed dynamically is very easy, we just need to flip a bit in the CR register back and forth. However, it might be the case that just the connection of the touchscreen controller causes trouble at this speed. We'll see. 
Then we have to decide which way to go. That is what experimental work is all about.

df8oe:
I have removed all touchscreen signal lines from SPI bus. Same result:  of the three LCDs show always "white screen" at bootup. It is possible if I dramatically warm up (45° - 50°) LCDs that 1 of 5 bootups show scrambled screen.

The other one shown 1 of five a white screen also and the rest 4 times a scrambled screen wih i is ununsable after a short time.

Additional I have tested all LCDs on another mcHF - same result. Difference in behaviour is located in LCD not in ui board.

What do you use for a kind of LCD?

vy 73
Andreas

db4ple:
I received a HY28B a few weeks before, ordered via a well known auction site :-). I also own a HY28A SPI display, made several years ago (2-3 years). HY28A does not run w/ 50 MHz, HY28B does. The linux driver for HY28A/B uses 32Mhz as default clock for HY28A and 48 Mhz for HY28B. I think, if we keep 50Mhz, this should be only be used optionally. For now I would think, we should init the SPI bus with 25Mhz on all displays, but keep the code to be able to experiment with it.

df8oe:
I think so, too. There are some possibilities to check:
- signal line termination
- usage of SPI bus line drivers
- usage of newer (and faster) LCDs (at SPI only a few lines to connect)

HY28 is not the best choice today. Many cheap LCDs which runs with Raspis are available at german distributors - I have some of them here :) I must have a look if the controllers are compatible or drivers can be adapted easily.

At this stage I think we push your code as actual devel state and add a #define for quickly change speed.

vy 73
Andreas

db4ple:
Agreed.

PR #147, opened 2016-03-11
Improved encoder handling of all encoders avoiding lost counts due to…

dl2fw:
… detent algorithm (detent++ changed to detent +=...) reduces lost Counts to a minimum

Optimized dynamic Tuning - using a Counter triggered inside the RX DMA sequence

should now be not influenced by Display mode timing


DD4WH:
100% improvement of the behaviour of all four encoders! Thanks very much, Michael!
73 de Frank DD4WH

PR #159, opened 2016-03-15
Changed order of menu buttons in order to reduce

db4ple:
risk of wrong reset to default.
Reorder some instructions so that display output calls are clustered.


df8oe:
I have added small changes for working completely.

vy 73
Andreas

PR #167, opened 2016-03-19
Simplified  Filter Path Definitions. Moved filter coeffs from .h to .c Part 1

db4ple:
Does not change any functionality, just cleanup.
Makefile and Eclipse work, CoIDE build info not yet updated.


df8oe:
Hmmmmm....

Does not build?!

make: **\* No rule to make target `fir_rx_decimate_4.o', needed by`mchf.elf'.  Stop.

I have not cheked that's the result of TravisCI. It has alarmed 4 times in the past and every alarm was correct.

vy 73
Andreas

db4ple:
I have to use travisCI to check if my makefile modifications work (which they did not initially :-( ).

PR #171, opened 2016-03-20
simplifying filter switching

DD4WH:
Tried to split:
1. filterbandwidth display: is in UiDriverChangeFilterDisplay and
2. filter switching/intitialising is in audio_driver_set_rx
unfortunately, now mcHF crashes every time when an IIR filter instance goes from NULL to another state!? --> eg. filter_path = 44


df8oe:
I do not merge this now because I do not have time today to look what happened.

vy 73
Andreas

PR #188, opened 2016-03-25
Slight change in visual appearance for filter info

db4ple:


df8oe:
We both have worked on exactly the same line of code :) Your pull request reached me while I was working on that - I have implemented manually - please rebase.

vy 73
Andreas

PR #189, opened 2016-03-25
New FIR filters 4k5 & proposal for graphical cleanup

DD4WH:
I was not able to read black figures on orange background --> changed to white figures.
Also cleaned up a bit --> no more grey top line which really is useless and does not look good.
Please test and discard this pull request, if you do not like the killing of the grey line.
But please no black figures on orange background ;-)


DD4WH:
Added Hilbert filters 4k5 and corresponding phase correction files. --> used for SSB & SAM
Added FIR filters 4k5 --> used for AM
Hopefully this improves reduction of heterodynes when using filter bandwidths from 3k8 to 4k8 when using SSB & SAM.
bugfix: Hilbert filter has to have -90 degrees phase shift, NOT +90 degrees, otherwise LSB/USB swapped again ;-)

df8oe:
I switched to black letters from white with good reasons! If you use mcHF outside in a little bit sunny place (and it is a portable transceiver!!) you could not  read anything if letters are white on orange background but very good if letters are black so it stays so. Even in darker surrounding reading of black letters is much better than of white ones. I cannot read white on orange...

vy 73
Andreas

PR #231, opened 2016-04-04
Fixing white screen for parallel LCD

phaethon:
This fixes parallel LCD white screen. UiLcdHy28_ReadByteSpi returns 0xff
on timeout (consequently UiLcdHy28_ReadDataSpi returns 0xffff), but
UiLcdHy28_InitA checks for return 0x0000

### DIFF
--- a/mchf-eclipse/drivers/ui/lcd/ui_lcd_hy28.c
+++ b/mchf-eclipse/drivers/ui/lcd/ui_lcd_hy28.c
@@ -468,7 +468,7 @@ static inline void UiLcdHy28_WriteDataOnly( unsigned short data)
 //*----------------------------------------------------------------------------
 unsigned short UiLcdHy28_ReadDataSpi(void)
 {
-   unsigned short value;
+   unsigned short value = 0;
    uchar y,z;
 
    GPIO_ResetBits(lcd_cs_pio, lcd_cs);
@@ -1018,7 +1018,7 @@ uchar UiLcdHy28_InitA(void)
    ts.DeviceCode = UiLcdHy28_ReadReg(0x0000);
    printf("lcd id: 0x%04x\n\r",ts.DeviceCode);
 
-   if(ts.DeviceCode == 0x0000)
+   if(ts.DeviceCode == 0x0000 || ts.DeviceCode == 0xffff)
       return 1;
 
    // HY28A - SPI interface only (ILI9320 controller)


df8oe:
Very astonishing. I do have 4 mcHFs for my own - two of them use LCD in parallel mode (so no SPI wires for LCD - only for touchscreen). Many of the users in our German project group use LCD in parallel mode and do not have LCD SPI wiring for LCD. None of those do have any problem with newer firmware.

There must be a change or specific charge of LCDs which react quite different to hundreds of other LCDs...

Many thanks for your work! We were not able to do this because no of our LCDs as malfunction.

vy 73
Andreas

phaethon:
Your are welcome ;)
Can you explain in the code how this worked for others? If UiLcdHy28_ReadByteSpi returns 0xFF on timeout, then it means there is no timeout for others and SPI_I2S_ReceiveData returns 0 (meaning SPIx->DR is 0), which is strange (at least no timeout with no SPI wires part).

df8oe:
After digging into logic explanations I am not sure this will fix the problem.

Why does he issue affect only specific LCDs with 100% reproducable result?
Many mcHFs are running in parallel without SPI wires and do have or do not have touchscreen modification (which also uses SPI bus) with original code.

But there exist some rare LCDs which shows a white screen - and this on ALL ui-boards.

We do have had a problem with reset timing a few weeks ago. I could solve it by adding a single code line to a function, which 100% NOTHING has to do with reset - but it solves problem for all. Next firmware release some other lines changed - there the "new line" mut be commented out and reset issue was gone again...

Means:
For now it works, but we must be carefully watching this - I hope with your help. I cannot help because I do not have access to a LCD with this issue.

Regarding your question:
Nothing changes for others. If it would be a problem to get randomly 0x00 or 0xff mcHFs do randomly show white screens regardless of LCD. But this not occurs. 99.99% of all LCDs work flawlessly.
If there would be a misidentification of LCD mode wrong initialization would occur on random LCDs. And if it is misidentified at start, it will RIP until you unplug power or switch mcHF off. That does not occur.

But we see white screen reproducable only with some rare LCDs - the change of the code you have done cannot explain that this was responsible :( I think it was a solution "by chance" and white screen will appear next time again (of course I do not hope that)...

vy73
Andreas

EDIT:
I have changed identification string in hardware menu some weks ago for identifiying controller chip on LCD instead of wiring (HY28A/HY28B). Please take a look at hardware menu and tell me the identified LCD controller chip - maybe this is interesting...

phaethon:
Can it be due to lack of pull up/down for the pin as Danilo mentions on Yahoo mg? Others get 0 signal on the mc pins, I get 1, without a pull down this can happen.
If I understand correctly, mchf code does not set internal pull up(down) through GPIOx->PUPDR. Can (should) we use it?

df8oe:
No, that cannot be the reason. In this case issue would be ui-board specific (flux, high resistant connections e.g.) and LCD independent.

But it is _not_ ui-board specific - it is LCD specific!

I know one other OM in Serbia who has two mcHFs with pluggable LCDs. One LCD shows issue 100% on both boards - other LCD works flawlessly also on both boards. And "issue-LCD" works for him with older firmware.

I have sent him new binary for testing - we will see...

vy 73
Andreas

EDIT:
To get light into the darkness you can check it this way:
- flash firmware which gets white screen again
  --> solder 2.2K from MISO to GND and check function
  --> solder 2.2K from MISO to Vcc of STM32F4 (3.3V) and check function

Feedback VERY welcome...

phaethon:
May be some LCDs are more 'noisy' and consequently unconnected mc pins without pullup/down are less stable when using those LCDs? Using pullup/down would make sense for me anyway. Currently, the code is reading a possibly unconnected pin and expecting it is zero. I will experiment in the evening with PUPDR and see if it makes difference.

df8oe:
Doing the tests with malfunctioning firmware and once external pullup and other pulldown will show this. And I am very interesed in results - because I cannot do this here...

df8oe:
Update:
I have sent binary to three users I know they have had the "white screen issue". All three gave feedback "it is working now". For all other users who never have had a problem nothing changes.

I have tested on one of my mcHFs running LCD in parallel:
- removed touchscreen modification so that SPI pins are floating.
  result: LCD is working.
- added 2.2k resistors from MOSI&MISO to GND
  result: LCD is working.
- changed resistors to +3.3V
  result: LCD is working.

So I can 100% state:
New built firmware does fix issue. But I don't understand why :(

Please can you perform the test with resistors to GND or +3.3V, too? This will help to decide if voltage on these pins do affect LCD function or not.

If LCD works in both cases for you, too reason cannot be different returns of SPI read...

Additinally it would be very nice if you can perform a test with firmware which gets hite screen and look if this is changeable by resistors...

phaethon:
At least I hope you will agree that I shouldn't hurry throwing out my LCD :)
I will try to do the tests with physical pullup/down. Also, I plan to investigate internal microcontroller pullup/down and its impact on the issue. But probably not today.
It would seem correct to have either physical or internal pullup/down for pins anyway, even if it works for many without it.
I actually don't understand how it can work for you without my code change and with a pullup to 3.3v. How can UiLcdHy28_ReadByteSpi not generate timeout? Can you debug and test that UiLcdHy28_ReadByteSpi actually returns zero even with a 3.3v pullup without wires attached? How can it happen?

df8oe:
Until now nobody of the 3...5 users who have had problems before has the issue again.But it were hundreds (!!) of users who never have had a white screen issue and that tells me that ost part of the problem is located in LCD. But I want to understand regardless other things.

PR #251, opened 2016-04-14
notch filter#1 not yet functional

DD4WH:
first trial for a manual notch filter. Functionality not changed otherwise.
Long press G3 activates the manual notch filter in the RX path (silence at the moment . . . )
TODO:
1.) I do not seem to get the init of the biquad IIR filter right . . .
2.) Adjustment of notch_frequency with encoder2 to be implemented
3.) separate void for the calculation of the notch IIR filter coefficients to be written
4.) menu entry --> Q = depth of notch (the higher the Q, the narrower the notch) to be implemented


DD4WH:
FM cleanup finished: deleted all treatment of FM bandwidths in the menu section --> also old EEPROM saving commented out

PR #297, opened 2016-04-23
Fixed the noise issue when tuning

db4ple:
A true classic, have a look at the diff in ui_si570.c and enjoy!


df8oe:
absolutely classic...

vy 73
Andreas

PR #303, opened 2016-04-23
proposal switching notch display box

DD4WH:
- Display box for notch filter disappears only when notch is switched off
- Filter choice by encoder 3 is default now
- Changed hard coded references in spectrum display code
  When trying to double the precision in the FFT code, RAM overflow was an error reported by the compiler (51kB too much)


db4ple:
I like the replacement of numbers with FFT_LEN etc. Good work!

PR #325, opened 2016-04-28
Tx Power Issue (Temporary)

db4ple:
Please Test, if does not chane situation, have a look at the two functions around the commented out function.


df8oe:
works.

vy 73
Andreas

PR #438, opened 2016-07-05
added 1k8 SAM filter & quick bugfix Tx compression level

DD4WH:
very difficult reception yesterday evening of Radio Candip, DR Congo, 5066.4kHz, 1kW tx power, so I had to design a new narrow SAM filter 1k8 . . . beautiful african music, but poor reception ;-)
quick bugfix for Tx compression level --> now contains 13 levels of compression.


db4ple:
Very  good. I'll make the code a bit more robust then but the added value pair was the missing piece.
73
Danilo

PR #445, opened 2016-07-12
Timeout for CAT buffer and number of small changes, most do not change anything

db4ple:
Changed cleanup approach  for CAT buffer to use  a timeout.
The 24C256 EEPROM is now being usable for data storage (again).
Most changes just clean up small things.

### DIFF
--- a/mchf-eclipse/drivers/audio/audio_driver.c
+++ b/mchf-eclipse/drivers/audio/audio_driver.c
@@ -214,26 +214,7 @@ void audio_driver_config_nco(void)
 //*----------------------------------------------------------------------------
 void audio_driver_init(void)
 {
-    ulong word_size;
-    uchar x;
-
-
-#ifdef DEBUG_BUILD
-    printf("audio driver init...\n\r");
-#endif
-
-    // "use" the temporary variables to keep the compiler from "optimizing" them out of existence
-    //
-
-//	test_a[0] = 0;
-//	test_b[0] = 0;
-//	test_c[0] = 0;
-//	test_d[0] = 0;
-//	test_e[0] = 0;
-//	test_f[0] = 0;
-//	test_j[0] = 0;
-
-    word_size = WORD_SIZE_16;
+    const uint32_t word_size = WORD_SIZE_16;
 
     // CW module init
     cw_gen_init();
@@ -251,9 +232,11 @@ void audio_driver_init(void)
     ads.agc_var = 1;			// used in AGC processing
     ads.agc_calc = 1;			// used in AGC processing
     ads.agc_holder = 1;			// initialize holder for AGC value
-    for(x = 0; x < BUFF_LEN; x++)	// initialize running buffer for AGC delay
+    for(uint32_t x = 0; x < BUFF_LEN; x++)	// initialize running buffer for AGC delay
+    {
         ads.agc_valbuf[x] = 1;
-    //
+    }
+
     ads.alc_val = 1;			// init TX audio auto-level-control (ALC)
     //
     ads.fm_sql_avg = 0;			// init FM squelch averaging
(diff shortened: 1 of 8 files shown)


Astralix:
db4ple, you can keep function return values, even they are not used for the moment. The compiler always reserves one CPU register for the first parameter and re-uses it for the return value, regardless if you use a return value or not. But it is helpful to have some information coming back for later optimizations. 
However it is funny that there where so much arguing around time-outs that require settings in the PC that are absolutely no-go. So I tried just to remove some leading bytes to get a multiple of 5 for the CAT and now this... 73 Astralix aka DC3AX

df8oe:
Hello Astralix,

there _is_ some trouble around timeouts...

I am using a Raspberry Pi running LXDE and self build WSJTX. Sometimes there is a big delay because of Raspi "has to do urgent things" - if it returns to WSJTX, the first USB packages are thrown away and the rest is running to innocence. If so, Raspi stops with a message box "CAT control failure..." - very nice if I want to do band hopping at night...

So timeout problems _are present_ and you cannot be 100% sure time for completing a package is < 300ms.

Solution is a mix of your implementation and the modification DB4PLE has one. At the moment my Raspi runs for five nights and there is not one CAT error...

73
Andreas

db4ple:
Hi,
I also use WSJT-X both on Linux and on Windows. WSJT-X is very sensitive when it comes to delays on the PC side. E.g. whenever I unlock the screen saver on linux I get an timeout shown on Wsjtx. For sure the problem is not in the mcHF since  if I run without screensaver I almost never get the timeout.
Never had a timeout  with HRD or fldigi, I think they simply retry. On the mcHF side, I never had issues with the 5 bytes not being in sync except if I forced it by sending incorrect numbers of bytes using a terminal program. And the old code (without sync) would not be able to recover from such a problem. So if it would be a common problem, it would have been reported. Anyway, a way to get rid of out of sync bytes is a good idea. With a command timeout programs which stop in the middle of sending a paket will be able to get a sync again if they wait long enough.  The modulo sync would  do the same with old data but unfortunately does not distinguish between old and new data. So if you send single bytes via USB (which works  for an FT-817) the sync would kill the bytes (as the driver is called often enough so that it would see each single USB frame and consequently the buffer filling up.
Such a program would not be able to work with the mcHF. I admit, haven't seen such a program but this does not mean, it does not exist.

73
Danilo

Astralix:
We should figure out if the serial communication is (mainly) done in full packets per USB frame, or if there are high chances of partial frames coming in by multiple USB frames. I guess that the software is sending out the communication in stings and leaves it to the driver to send the physical data at its will. So I bet on high chances that one USB frame carries one or multiple frames of 5 bytes. 
However, I uninstalled several drivers and still seem to have something inspecting the mcHF on startup and that still kills any CAT on the mcHF. I don't want users to spend hours to configure their PC to just work with mcHF and I don't want to buy another PC just exclusively for mcHF use. So I like to have the CAT packet handling to be fail safe.
As fast frequency setting results in a crashing SI570, we should force some timeouts. And if we force some timeouts, we can use these to re-sync the packets.
For me the simple fix is the current solution, using timeouts and cleaning out leading bytes. It needs some explanation in the Wiki as people should set timeout, delay and retries in their hamlib settings.
I know that timing is critical in mcHF but adding more time critical things on top wouldn't fix it.
The only alternative would be to chose any other CAT protocol that allows some sort of frame checking.

df8oe:
Running Linux there are mainly two applications which sends unwanted commands without asking the user. These two applications ust be deinstalled (or the authorsshould be informed to change this behaviour):
- modem-manager
- brltty

You can blacklist mcHF USB for modem-manager, so only brltty remains...

I have tested mcHF CAT using different "clean" Linux installations and all worked out of the box (some needs removing/disabling modem-manager and/or removing brltty). But that was all: no other problems occured.

Changing timeout inapplications is critical because of delaycan result behind these settings. Raspberry runs in timeoutsbecause of delay is added in packet handling after packets leave WSJTX (or other applications). It is possible that there is a bigger delay regardless of setting in hamlib.

73
Andreas

PR #458, opened 2016-07-15
snprintf does not seem to handle too long strings correctly. 

db4ple:
Increased string length. Did not test, but it is my only hunch after briefly looking at the vsnprinft code.
Just try and  see if it does not crash the Sys Info Menu,  since with certain messages we did break the 32 character size limit.


df8oe:
Hallo Danilo,

ich merge das jetzt nicht - schau bitte mal in unser Forum...

PR #469, opened 2016-07-25
Added new  TX USB Audio out debug option, simpler temp convert 

db4ple:
2 unrelated changes.
Temp convert algorithm changed -> no functional change
It is now possible to get TX IQ data  via USB Audio Out. Enabled via new debug menu. During RX USB Audio Out is as usual. Otherwise no functional change


ly2bok:
```
Added debug option to route digital IQ audio to USB when transmitting......
```

This is very fine, but what and where is regulated?
LY2BOK

DD4WH:
as stated, that is a debug option for the developer. If you do not know what to do with it, just ignore it. It will not change the functionality of your mcHF.
If you would like to use it for debugging, please be a little more specific in your question(s).
73 de Frank

PR #470, opened 2016-07-26
TX filter & software DDS spur issue fixed, but still problems

DD4WH:
Software DDS altered, now in SSB tune without spurs! 
BUT: 
- CW Tune does not work
- double tone TUNE does not work
  TX filters functional, but still not optimal
  Bass & Treble TX adjustment in preparation, but not yet functional


df8oe:
I merge follow up #471

PR #478, opened 2016-07-28
Replaced the 0 degree Hilbert Filter with a simple delay function

db4ple:
Still uses the "normal" data structure of the filter to buffer data.
This is inefficient memory-wise, if it works out, will make that using much less memory.
Not yet tested what effect it has on the signal.


df8oe:
This version is buggy. If I try to TUNE or transmit mcHF switches OFF.

PR #578, opened 2016-09-28
FreeDV now uses arm*fft routines to reduce memory footprint

db4ple:
Moved to use of codec2_fft wrapper and added one more use of real fft.
Removed a few more of local stack arrays. Switched to use arm_*fft for
FreeDV. This gives a little gain in speed but huge reduction in memory
usage.


df8oe:
Hi Danilo,

as you told me I shall not merge if build fails...

vy 73
Andreas

PR #581, opened 2016-10-01
More Power for FreeDV. Use with care! Not yet calibrated.

db4ple:


df8oe:
We have worked both on same lines :) Now it is resting at "20".

db4ple:
Fine with me :-)

PR #582, opened 2016-10-01
added compile time option MCHF_BOARD_0P5 to enable TP on v0.5 board.

shkmr:
This change enables touch screen for v0.5 board when compiled
with -DMCHF_BOARD_0P5.  For example

   $ make EXTRACFLAGS=-DMCHF_BOARD_0P5 clean all

will create mchf.bin for 0.5 board.

### DIFF
--- a/mchf-eclipse/hardware/mchf_board.h
+++ b/mchf-eclipse/hardware/mchf_board.h
@@ -105,8 +105,13 @@
 #define ADC2_RET_SOURCE		GPIO_PinSource3
 #define ADC2_RET_PIO       	GPIOA
 // pin 4
+#ifdef MCHF_BOARD_0P5
+#define TP_IRQ			GPIO_Pin_14
+#define TP_IRQ_SOURCE		GPIO_PinSource14
+#else
 #define TP_IRQ			GPIO_Pin_4
 #define TP_IRQ_SOURCE		GPIO_PinSource4
+#endif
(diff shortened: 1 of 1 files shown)


shkmr:
I just wanted to try new FreeDV mode.  But I could not find a way to enable it without touch screen.

db4ple:
You have to uncomment the last two lines of void  FreeDV_mcHF_init() in freedv_mchf.c
Now the mcHF starts in FreeDV mode. The "digital" box will not show FreeDV but the flickering green led tells you that you're in FreeDV mode.

We'll make that a little more convenient later on.

df8oe:
If you do not want to build every release by yourself it is recommended you add wires like signal is used on 0.4 boards. It was a very poor idea to cut off live debugging lines in 0.5

shkmr:
db4ple,  thanks for the info.  I was about to ask if this flickering green led is normal.
df8oe, thanks for the advice.  I already have st-link and planning to make the mod.
Still, it is nice to see just changing port assignment is enough for unmodified 0.5 board :-)

df8oe:
Of course you can use that and it is working well. But we do not want to offer different firmware releases for 0.4 and 0.5 so it is a _must_ that important hardware like touchscreen uses same ports on both boards. But your code improvement does not affect normal builds and make "special builds" very easy so I have merged it of course.

PR #587, opened 2016-10-03
fixed font 4 issue - it was the width correction which was to big on …

dl2fw:
…some characters, so it was interferring with the previous character...

added some freedv message functions
switched to font 4 for txt message - srolling txt follows...


db4ple:
Hi Michael,

please fix the Width function calculation as well. Unfortunately you'll have to apply your changes twice. 

Thank you!
Will unify both later. But not while you're working on it.

PR #600, opened 2016-10-14
Added FreeDV scrolling text message

dl2fw:
... will modify the font - there are still varying gaps...

### DIFF
--- a/mchf-eclipse/drivers/audio/freedv_mchf.c
+++ b/mchf-eclipse/drivers/audio/freedv_mchf.c
@@ -461,9 +461,7 @@ void  FreeDV_mcHF_init()
 
     f_FREEDV = freedv_open(FREEDV_MODE_1600);
 
-
     sprintf(my_cb_state.tx_str, "CQ CQ CQ mcHF SDR with integrated FreeDV codec calling!");
-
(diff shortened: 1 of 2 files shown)


mchf-eclipse/drivers/audio/freedv_mchf.c
```diff
@@ -199,9 +203,11 @@ void fdv_print_ber()
 }
 
 void fdv_clear_display()
-{
+{ char clr_string[freedv_rx_buffer_max];
+
+  snprintf(clr_string,freedv_rx_buffer_max,"                                                                           ");
```
shkmr, line 208: why not to use memset?

db4ple: Maybe because some likes to count spaces, since it is a hard and difficult task?
Maybe because it did the job?
Maybe because you didn't sent a pull request with the change :-)

dl2fw: Hi, of course something like memset would be much more efficient and straight forward..
The reason why it is still like that is, that I'm working on the intermediate font spacing, txt position and length, and am using the line of spaces for this work. In that case I'm copying the characters I'm currently working on to that line and can directly see the effect rather than working with received text...

PR #610, opened 2016-11-08
Added gnuradio Simulation of mcHF

db4ple:
and did some other rather small changes, see commit logs please


AngelSerra:
Add a new band of amateur band of 11 meters 40 channels

AngelSerra:
Transmit cb27 band, 11 meters 40 channels

df8oe:
That is no comment for a commit! If you do have feature wish, please use ISSUES.

All comments what are not related to a commit will be DELETED.

Issue #233, opened 2016-04-06
Change storage  Image rejection for mode to Image rejection for frequency


PA3CCE:
I like the new IQ balance adjustment,
now I can turn the image completely into the noise !

In the menu, storing of the IQ balance settings is provided
for the different modes LSB, USB etc.

In my case, I find no- or neglectable differences in
settings between the modes, but quite some difference
in the settings over the entire frequency range.
If confirmed, I should like to have the IQ balance in the
menu per mode, changed to settings per frequency
(Low-, medium-, High Frequency, or even better per band).

Jos (PA3CCE)

df8oe:
Hello Jos,

this is a very good idea. When I first adjusted IQ with the new functions I had to increase power because I could not measure unwanted signal! And so I stated that there is a dependency to power, too. If you adjust it @0.5W and signal is completely gone, you switch to 5W and adjust that signal again is completely gone, then switch back to 0.5W you will see there is a slight misadjustment.

But we are talking about things which before nobody was able to see :)

We will think about improved structure, freuency dependency included. It is not high priority because now its working for the first time, but "good things can be improved" of course :)

vy 73
Andreas

PA3CCE:
Well, I don't find the present situation real "good", when set at mid frequency,
the image rejction at low- and high-frequencies are rather "poor"

73 de
Jos

df8oe:
...but _much_ better than all the years before :) So I call it 100% "good". Hundreds of users have lived without a working IQ-adjustment and nearly nobody has complained before. But now, if it works - it is _not good_ ??

Sorry - I cannot understand.

vy 73
Andreas

PA3CCE:
A change from not working at all to 100% working for just one frequency 
is huge.
Working for the entire frequency range is  much better.
I think that many users just calibrate for one frequency and think " OK 
fine".
They  seem not to be aware what happens at  other frequencies.
Besides, I never heard you say before that the image rejection was not 
working at all,
or did I miss that ?

Best 73 de
Jos

Op 8-4-2016 om 11:06 schreef eeDF8OE:

df8oe:
Hello  Jos,

haven't you read many posts in the past in Yahoo-NG that IQ-phase adjust does not work?? These posts go back to the time before I joined the project in May 2015... So the best adjustment in the past was not so good as if you adjust it with actual firmware and go to the band with the worst rejection. Because of it was not working properly before many users never have had adjusted their iq settings.

[...]

PR #633, opened 2016-12-04
Highly Experimental: Frequency-based IQ Phase and Gain Balance

db4ple:
This is in reference to #233  	

Adjust IQ Phase and Gain Balance at low (around 3600khz) and high (around 28100khz)
frequency using USB.

THIS IS HIGHLY EXPERIMENTAL! MAY NOT WORK AT ALL!
Needs external validation!


df8oe:
I will merge this because it is the future of better IQ adjustments. May be it needs more working on it - we will see...

PR #637, opened 2016-12-07
Fix for IQ settings problem on TX

db4ple:
Plus small refactorings (renamings in ui_spectrum.c and resulting changes in some other files)
No functional change except for issue fix in audio_driver.c (single line change)


df8oe:
baut leider nicht...

PR #708, opened 2017-01-03
first attempt one sideband SAM demodulation, not yet working

DD4WH:
In fact, it is working ;-), but the MCU is not fast enough to deliver nice audio ;-). We have to build another audio path, maybe like this:
48ksps - Freq_IQ_conversion - decimation I & Q separately - 12ksps - SAM demodulator - interpolation - 48ksps

### DIFF
--- a/mchf-eclipse/drivers/audio/audio_driver.h
+++ b/mchf-eclipse/drivers/audio/audio_driver.h
@@ -64,6 +64,7 @@
 #define SCALING_FACTOR_IQ_AMPLITUDE_ADJUST 2731.0
 
 
+
 typedef struct
 {
     // Stereo buffers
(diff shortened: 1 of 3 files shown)


mchf-eclipse/drivers/audio/audio_driver.h
```diff
@@ -184,6 +185,8 @@ typedef struct AudioDriverState
     float32_t               onem_mtauR;
     float32_t               mtauI; //(exp(- DF / (IQ_SAMPLE_RATE_F * tauI))); //0.99999255955;
     float32_t               onem_mtauI;
+    float32_t               c0[7];          // Filter coefficients - path 0
```
db4ple: please replace 7 with STAGES 
probably STAGES needs to be defined in the header then

DD4WH: OK, thanks for the hint!
DONE!

Issue #773, opened 2017-02-04
DTR for CW carier


WD8BXS:
Wow, this looks like a great add, I do not get it to work in fldigi, Where exactly do I set it up in the fldigi?
I have selected DTR for PTT but no luch.

Thanks,
 de Chuck WD8BXS

db4ple:
Hi,
it is NOT PTT via DTR it is CW Key via DTR, PTT is still only FT-817 CAT. The DTR line can be used like a straight key in order to enable / disable carrier (when TX was enabled by CAT PTT command).

WD8BXS:
I do not understand, Is this something that can be used in fldigi to send CW with the rig in CW mode? Here is a pic of my screen, I see no where to address what yo are talking about.
![rigcat](https://cloud.githubusercontent.com/assets/25059155/22618901/a0f0e5f4-eab6-11e6-9f17-0f01de4ce2d9.jpg)

db4ple:
I have no idea how fldigi does CW. I have also no clue how rigcat works, never used it, only Hamlib. 
The request is based on the problem to get UCXLOG to work, not fldigi. 

BTW, in the current daily this function does not work at all correctly.

df8oe:
It is not a neccessary feature for most external software. Normally you can control mcHF via CAT commands (this works for SSB and CW). The best choice is to use "only CAT commands".

Helpers like "PTT via RTS" or "DTR" are from many years ago when CAT does not support switching PTT or doing CW via CAT.

73 de
Andreas

dm6tt:
As much as I know fldigi uses the right audio channel with a 1600Hz tone to do cw-keying.
http://www.w1hkj.com/FldigiHelp-3.21/html/cw_keying_page.html

DG9BFC:
ptt via rts would be a nice add on for digital mdes software that can not cntrol ptt via cat command
then any software that can toggle rts can be used with mchf

df8oe:
If it is ok now please close this issue.

PR #774, opened 2017-02-04
CW via DTR: Fix for previous pull request. Now working as expected.

db4ple:
If in CW Mode, doing PTT via FT817 CAT (!!!) will allow to
key the transmitter using the (virtual) DTR of the CAT serial port.
This has been tested with UCXLOG and MAY work with other programs.

See also #773 for a little bit of discussion

### DIFF
--- a/mchf-eclipse/.settings/language.settings.xml
+++ b/mchf-eclipse/.settings/language.settings.xml
@@ -5,7 +5,7 @@
 			<provider copy-of="extension" id="org.eclipse.cdt.ui.UserLanguageSettingsProvider"/>
 			<provider-reference id="org.eclipse.cdt.core.ReferencedProjectsLanguageSettingsProvider" ref="shared-provider"/>
 			<provider-reference id="org.eclipse.cdt.managedbuilder.core.MBSLanguageSettingsProvider" ref="shared-provider"/>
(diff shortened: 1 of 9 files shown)


WD8BXS:
I Installed the UCXLOG, and I can't even get it to copy CW, it keeps telling me that there is no output sound card found

db4ple:
What can I say? I have installed UCXLog and it worked. 

Here is my "Other Interfaces"  Station Settings: 

image

And for reference the "Transceivers" tab as well:

image


BTW, I have never used UCXLOG for anything but testing the CW output, so it seems I am just lucky to get it going without problems.

WD8BXS:
OK, I got it to send CW with the MCHF,  all is well there, I did not realize that it did not copy as well , my fault there.Interesting to say the least, but not to my fancy. I will stick to the fldigi and my Aclog program hihi.  maybe fldigi will catch up and allow it to key the radio for CW and not just use audio SSB.

db4ple:
UCXLOG is user interface wise something outside of what I would consider "normal" but it seems to have its share of users.

Re fldigi: it is open source, so shouldn't be too hard to get it doing what one wants....

BTW, anyone is invited to provide the QSK code for the mcHF. It is not hard either, we discussed this in the german forum as an option if the DTR approach wouldn't work out as expected.

PR #775, opened 2017-02-06
Changed frequency calibration to PPM based corrections

db4ple:
Switched to use float64_t for frequency inside ui_si570.c to improve precision
We now stored a ppm correction value which is applied to the base fxtal of the Si570


df8oe:
I will test tomorrow morning and release binary if it is ok.

df8oe:
works perfectly for me.

PR #808, opened 2017-02-17
Active devel

AsbjornPettersen:


db4ple (commented):
Hi,

no. This variable does need to be initialized. All variables not initialized are 0. Which is okay.

PR #814, opened 2017-02-18
Add Microsoft C defines

AsbjornPettersen:
Add Microsoft C defines.
Prepare unit tests with the MSC compiler

### DIFF
--- a/mchf-eclipse/cmsis/arm_math.h
+++ b/mchf-eclipse/cmsis/arm_math.h
@@ -439,6 +439,9 @@ typedef double float64_t;
 #define __SIMD32_TYPE __unaligned int32_t
 #define CMSIS_UNUSED
 
+#elif defined ( _MSC_VER )
+#define __SIMD32_TYPE int32_t
+
 #else
 #error Unknown compiler
 #endif
--- a/mchf-eclipse/cmsis/core_cm4.h
+++ b/mchf-eclipse/cmsis/core_cm4.h
@@ -115,6 +115,10 @@ extern "C" {
 #define __INLINE         inline                                    /*!< inline keyword for COSMIC Compiler. Use -pc99 on compile line */
 #define __STATIC_INLINE  static inline
 
+#elif defined ( _MSC_VER )
+#define __INLINE         inline  
+#define __STATIC_INLINE  static inline
+
 #else
 #error Unknown compiler
 #endif


db4ple:
Hi,
could you please outline what you intend to do and why we should put your pull request into the main repository?

In general, we expect a contribution to be complete in the sense of "it makes sense to give this to the other users" (for testing, for experiments etc.) but a pull request should have a clear benefit. 
As long as this is not given, the strategy is to work in private branches (and to do regular "git rebase" as per instructions in the mcHF Wiki in order to keep the final code as easily mergable as possible).

If you need help with this, let us know. We really appreciate contributions, but please follow the "protocol" which helps us to keep the firmware working. For larger contribution it is best to announce the intent to do something in the issues section (as an issue), let us have some discussion if necessary, but then everyone involved knows what is going on.

Thanks,
Danilo

df8oe:
We don't want to switch to a closed source compiler - or what is the intention to add this? As Danilo already stated - I do not see a benefit here...

Please explain what you want to do. Thank you for starting contribution!

73 de
Andreas, DF8OE

AsbjornPettersen:
My plan is to use the MSC compiler for making unittest code, and also to learn about of the codebase.
The MS Visual studio is a great tool for investigating the code, but i havn't investegated the support for ARM yet.
I'm also using the gcc for ARM to see that the code builds.

The ( _MSC_VER ) will get the code build, but have no real value yet.

Planning to buy be a mcHF card.
The unit tests without HW will not be added to the source base.

I'm now investigating the "hardware/mchf_board.c" code and i see some "copy and paste" code.
Here are some improved test code that could be used in mchf_board.c :
//
// this function will reduse the lines og code with approx. 10 percent
//
static void my_GPIOinit(GPIO_TypeDef *GPIOx, const uint32_t GPIO_Pin, const GPIOMode_TypeDef GPIO_Mode,
	const GPIOOType_TypeDef GPIO_OType, const GPIOSpeed_TypeDef GPIO_Speed, const GPIOPuPd_TypeDef GPIO_PuPd)
{
  GPIO_InitTypeDef GPIO_InitStructure;
  GPIO_StructInit(&GPIO_InitStructure); //overkill since we set it later

  GPIO_InitStructure.GPIO_Mode = GPIO_Mode;
  GPIO_InitStructure.GPIO_OType = GPIO_OType;
  GPIO_InitStructure.GPIO_Speed = GPIO_Speed;
  GPIO_InitStructure.GPIO_PuPd = GPIO_PuPd;
	
  GPIO_InitStructure.GPIO_Pin = GPIO_Pin;
  GPIO_Init(GPIOx, &GPIO_InitStructure);
}

static void my_init()
{
  // example
  my_GPIOinit(PTT_CNTR_PIO, PTT_CNTR, GPIO_Mode_OUT, GPIO_OType_PP, GPIO_Speed_2MHz, GPIO_PuPd_DOWN);
}

Rgds
Asbjorn

db4ple:
Hi,
unit tests which can be shared make sense, however, since you do not plan to release these, please do not create pull requests. You can and should keep track of your changes in your repositories (the local one on your disk and your github one). This repository here is the main public repository and should not be used to record local changes (there is no need to, this is very different from SVN due to distribution). 

Regarding the optimization of init code: In general we are in favor of code simplification and reuse by making functions but in the particular case, please leave it as is, since it will go away in future anyway. We're in the process of replacing the low level abstractions by STM32 HAL plus generated code from  STMCubeMX. So any optimization work in this are may be a interesting challenge but overall a waste of time.

BTW, another rule is to test the code before you contribute (unless you explicitly mention it not being tested)

73
Danilo

df8oe:
Hi Asbjorn,

I have had choosen github and not svn because github works in offline mode, too (what svn is not able to)... The way github works is different from the way svn does. So you can work on your own fork, add changes we have made an this "master branch" and test everything you want to do. Because of actual changes are not interesting for the master branch there is no pull request needed.

I am running Linux so I cannot help you with gui usage (only git on command line) - but some of us are working using Windows. So if you have questions you can ask us using github issues.

73 de
Andreas

PR #856, opened 2017-04-23
reorder data items

AsbjornPettersen:
C++ require that data items are in sequence, so the data table now match(in same order) of the struct.
Works in C and C++,  and change should have no impact of the ARM code.   I'm making unit test programs in c++ with Google test tool.  
 
init. of the struct must be in order !
{
                .size = 0,
                .supported = false,
                .pagesize = 0,
                .name = "No EEPROM"
},

### DIFF
--- a/mchf-eclipse/misc/serial_eeprom.h
+++ b/mchf-eclipse/misc/serial_eeprom.h
@@ -18,8 +18,8 @@
 
 typedef struct {
     uint32_t size; // in Bytes
-    uint16_t pagesize; // in Bytes
     bool supported; // i.e. big enough to be used
+	uint16_t pagesize; // in Bytes
     const char* name;
 } SerialEEPROM_EEPROMTypeDescriptor;
 


db4ple:
Hi,
you are partially right, g++ accepts designated initializers if and only if they initializers are starting with the first member (size) and all following initializers are direct successors in the structurce member order (they call it trivial designated initializers). Some other C++ compilers may not even accept trivial designated initializers. And some may even accept also non-trivial ones. All of this is out of  any current C++  standard AFAIK. 

And more importantly, we are not using C++ but C and here the existing order is perfectly legal and supported.  And if we ever would switch to C++, the compiler would tell us and before this it is a huge effort to spot these issues. And later fairly easy to correct. 

Since this is an very simple change, I think we can pull that in. But be aware, that changes for the sake of change with no clear benefit and potential negative impact may not get through so easily. 

Sidenote: 
The whole topic of designated initializers is under current discussion in the C++ standard community:
https://jamboree.github.io/designator-draft.html , http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0329r0.pdf  
But nothing has been decided for C++17: https://en.wikipedia.org/wiki/C%2B%2B17

AsbjornPettersen:
This is the first time i have seen this problem, but i like portable code, C or C++.
There is no idea to switch to C++, but i have made a unit test program with Google test (c++).
Tried to build the same test program in Visual studio 2015, but it rejected the ".size" syntax. But i was able to build it, and run it.  I found other problems in the C code, and it must be some "design" error. I'll create another issue for that, when i  understand the code.

df8oe:
I will merge this - but:
as Danilo I do  not see benefits, new features or the fix of an existing bug. I would be thankful if commits follow one of these three arguments otherwise work on firmware can grow immense and eating time and energy without stepping forward... This commit is in nature "philosophical"...

PR #861, opened 2017-05-02
use single screen init function

AsbjornPettersen:
Found duplicate screen init code in bootloader. 
The screen init will now be in 1 function and it's same for bootloader and mcHF.
Reduce the lines of C code, but bootloader binary will almost be the same.


db4ple:
Well spotted. Did you check that code size of bootloader does not increase (a few bytes are okay)?
If so -> +1, otherwise we'll keep this on hold.

AsbjornPettersen:
Did a test on linux, the bootloader size was reduced with 4 bytes. From 31806 to 31802

PR #863, opened 2017-05-06
Add EmBitz directory and project files

f4fhh:
Hi Andreas

This is an addition to the mchf project that allows to compile the main firmware and the bootloader using the EmBitz IDE on Windows. There is no change to the mchf-eclipse directory. The EmBitz project files are all in a separate directory named mchf-embitz. All of the building process, creation of binaries and libraries take place in this directory. See the Readme that I have included at the root of mchf-embitz.

Best 73
Nicolas F4FHH

### DIFF
--- a/mchf-embitz/.gitignore
+++ b/mchf-embitz/.gitignore
@@ -1,0 +1,4 @@
+*.depend
+*.elay
+/obj/
+/bin/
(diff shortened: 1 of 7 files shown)


df8oe:
Nice idea and much welcome.

PR #873, opened 2017-05-12
bugfix FM & alternative coeffs for sideband suppression

DD4WH:
- small bugfix FM
- alternative coeffs --> will test them in the future, maybe they have same opposite sideband suppression AND much better BASS response


mchf-eclipse/drivers/audio/audio_driver.c
```diff
@@ -3710,7 +3710,7 @@ static void AudioDriver_RxProcessor(AudioSample_t * const src, AudioSample_t * c
             //
             if(dmod_mode != DEMOD_SAM && dmod_mode != DEMOD_AM) // || ads.sam_sideband == 0) // for SAM & one sideband, leave out this processor-intense filter
             {
-                if(ts.filter_path < 48)
+                if(ts.filter_path < 48 && dmod_mode != DEMOD_FM)
```
db4ple: Why only apply this for filters below 48? Reason for asking: Any new or deleted filter (or reordering) would break that condition. So the question is what this the difference for filters lower than 48 compared to the ones above we could use or introduce in the filter description?

DD4WH: Hi Danilo,
yes, you are absolutely right. That was not well done from my side.
I would put it like this:
                if(FilterPathInfo[ts.filter_path].FIR_I_coeff_file == i_rx_new_coeffs && dmod_mode != DEMOD_FM)
Reason is that we need to decimate-by-4 before applying the Hilberts, if we use the new coefficients (all SSB & CW filters below 3k8) and we decimate AFTER the Hilberts, if we use the old coeffs. 
Thanks!
73 de Frank

df8oe:
I just returned from Alan Parsons concert in Oldenburg and quickly merged & built this nice commit ;) Merging was "yesterday" - and building (a few minutes later) is TODAY...

DG9BFC:
how cool is that?? alan parsons in concert?!?
if i had known that earlier :-)

PR #911, opened 2017-06-18
CW timing too fast

pa7n:
Timing CW was too fast. A dit in 12 WPM should be 100ms.
On my mcHF it was 86ms.

### DIFF
--- a/mchf-eclipse/drivers/audio/cw/cw_gen.c
+++ b/mchf-eclipse/drivers/audio/cw/cw_gen.c
@@ -226,7 +226,7 @@ static const float sm_table[CW_SMOOTH_TBL_SIZE] =
 
 void CwGen_SetSpeed()
 {
-    ps.dit_time         = 1650/ts.keyer_speed;      //100;
+    ps.dit_time         = 1920/ts.keyer_speed;      //100;
 }
 
 static void CwGen_SetBreakTime()
--- a/mchf-eclipse/drivers/audio/cw/cw_gen.c
+++ b/mchf-eclipse/drivers/audio/cw/cw_gen.c
@@ -68,8 +68,10 @@ typedef struct PaddleState
     ulong   port_state;
     ulong   cw_state;
 
-    // Smallest element duration
+    // Element duration
     ulong   dit_time;
+    ulong   dah_time;
+    ulong   pause_time;
 
     // Timers
     ulong   key_timer;
@@ -226,7 +228,9 @@ static const float sm_table[CW_SMOOTH_TBL_SIZE] =
 
 void CwGen_SetSpeed()
 {
-    ps.dit_time         = 1920/ts.keyer_speed;      //100;
+    ps.dit_time         = 1800/ts.keyer_speed+9;  // +9 =  6ms * 1/1500 =  0,006*1500
+    ps.dah_time         = 5400/ts.keyer_speed+9;  // +9 =  6ms * 1/1500 =  0,006*1500
+    ps.pause_time       = 1800/ts.keyer_speed-9;  // -9 = -6ms * 1/1500 = -0,006*1500
 }
 
 static void CwGen_SetBreakTime()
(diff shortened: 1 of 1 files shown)


mchf-eclipse/drivers/audio/cw/cw_gen.c
```diff
@@ -226,7 +226,7 @@ static const float sm_table[CW_SMOOTH_TBL_SIZE] =
 
 void CwGen_SetSpeed()
 {
-    ps.dit_time         = 1650/ts.keyer_speed;      //100;
+    ps.dit_time         = 1920/ts.keyer_speed;      //100;
 }
```
db4ple: Hi,
appreciate the contribution! And after taking out my calculator I figured 1650 is not the right value (too short pulse). However, 1920 isn't either (too long in theory). It should be 1800: 1500 ticks/s * 1,2 = 1800 = 1WPM. I can see how you calculated 1920 (1650 * 100/86). Have you measured other WPMs (especially slow ones)? And also the pause should be checked, i.e.  a dit and a pause together should be 200ms at 12WPM.  The slight difference between 1920 and 1800 could be due to the "rounding of the edges" the code is doing.

pa7n: I have checked with my oscilloscope and measured 100ms at 12 WPM and 50ms at 24 WPM. Maybe my MCU is running faster? Tomorrow I will check the slow speeds.
73, Erwin

db4ple: It is nearly impossible that your device runs that much faster. In this case things like PC USB connections would not work for instance. Did you measure a pause as well?

pa7n: This is what I measure with the 1920 value:
6 WPM: 208 ms pause: 224 ms
12 WPM: 100 ms pause: 115 ms
24 WPM: 48 ms pause: 61.6 ms
48 WPM: 21.3 ms pause: 35.2 ms

So, still some work to do here to get the timings right.

db4ple: The problem is that the algorithm adds some ticks (a tick is 1s/1500)  to the pauses (you can see that in your measurements: the higher the WPM the longer the pause).  Still, with 1800 instead of 1920 we would have fairly correct dit + pause times (in sum):
-> ((1800/1920) * (208+224))/2 = 202.5 ms, 100.8 ms, ...   
 


  The problem is that the algorithm adds some ticks (a tick is 1s/1500)  to the pauses (you can see that in your measurements: the higher the WPM the longer the pause).  Still, with 1800 instead of 1920 we would have fairly correct dit + pause times (in sum):
-> ((1800/1920) * (208+224))/2 = 202.5 ms, 100.8 ms, 51.3 ,  26.5
I saw there is a 2 to 4 tick delay in the algorithm (which probably accounts for the 0.8 - 2.5ms additional time)  AND there is the ramp up and down of the edges which probably lead to seemingly shorter dits (and dahs).

pa7n: But do you get the same results? If al mcHF run at the same speed we could compensate for timing differences depending on WPM speed, during transmit and during a pause. 
I think we need some correction for both transmit and during a pause. One value (1800?) is not enough to get the timing right.

db4ple: Can't measure right now. But I think your measurements make sense when I look at the code.
I must admit, I don't see this whole thing a major issue but if we change something in this area, we should do it right.

pa7n: This formula for calculating key on time is probably correct (I still have to test it):

ps.dit_time = 1800/ts.keyer_speed+9; // +9 = 6ms / (1/1500) = 0,006*1500

The 6 ms is the rise and fall time combined of the CW smoothing.
I will test it tonight (I'm not near my mcHF now). I will create a new formula to calculate the pause time. 
(A wild guess would be: pause_time = 1800/ts.keyer_speed-9; )

pa7n:
This could be the final fix. Timings are now perfect.

PR #915, opened 2017-06-22
Fixed build issue with newer compilers

db4ple:
changing from scalar values to struct ( mchf_waterfall ) caused -O3
builds to grow by ~64k. Not sure why, probably the interaction with volatile.


df8oe:
This is already implemented this in my last commit...

Issue #859, opened 2017-05-01
Enhancements for portable use


om1aeg:
Hallo, after SOTA & OMFF activation with mchf trx powered from 3S LiPo Battery i have idea for LOW BATTERY warning (configurable voltage level with some delay, sound or text on LCD) and switch off at configurable voltage level, also with delay.
Is it possible add this feature?

Also readability of LCD on direct sunlight is bad, maybe inverted colour schema or switchable normal / remove / minimize some items on LCD together with bigger frequency and S-Meter can helps in portable, Maybe dynamic LCD with 2 screens, one "basic" when touching LCD, knobs, buttons and one after few seconds when no "touch something on radio" with bigger frequency, S-Meter, waterfall, but without "level" , "mode" "input" items. 

Configurable lock is another enhancement for portable use.

Thanks, Anton, OM1AEG

df8oe:
Actually there is nothing planned regarding this. If you use 3.2" LCD and adaptor PCB brightness and contrast are better than on 2.8" LCD.

yo2ldk:
Referring to display, can be added waterfall + spectrum option, or this overload the STM ?

df8oe:
Overload of CPU and too low pixel amount for a useful view.

yo2ldk:
and if we use SPI LCD mode to free I/O, this will be possible?
https://www.seanet.com/~karllunt/S4DVGA.html

DG9BFC:
i think a user setable low batt or "undervoltage" alarm should be easy
and would be nice for users that have a built in battery pack
maybe "blinking backlight" as prewarning??
and/or a text that pops up (low battery alarm switching off in 10...9...8...7)???

...
we already have the voltage reading on the screen ... you just have to look on it when you are doing portable ... (but no preventing of deep discharge of a li ion pack!!!)

phaethon:
+1 vote for undervoltage alarm/shutdown with two level configurable thresholds.

phaethon:
I have created a simple implementation, which:
- adds configuration setting "Low power shutdown" under Configuration menu
- setting variates OFF - 9.0, 9.1, ..., 12.0V
- if not off it shuts down MCHF if voltage is below the threshold

Implementation is very similar to existing LCD blanking timer logic.

Adding a delay and prior notification could be useful for someone, but I found that I am often operating with LCD off when from battery. I would not notice visual notification. Adding a notification sound might be disputable for others.

[...]

PR #920, opened 2017-06-27
Low voltage shutdown from #859

phaethon:
Configuration of threshold and shutdown when voltage is below threshold as discussed in #859


db4ple:
Hi, thanks for the nice contribution. 

I have some comments though:

The purpose of the enable bit is not clear. In your implementation "OFF" is represented by 0, any other value is considered always to be an enabled value. 
So the enable bit is redundant. Unless you make this an independent setting.
In this case one menu entry is used to set the shutdownvalue, and another one sets the ON/OFF bit. So you don't loose the set value even if you disabled that temporarily. This would also allow to set a good default value for power off with power off disabled by default.

The other idea is to use at least seven bits to encode the voltage (127 -> 12.7 Volts) this permits to go much lower than 9V).  You might wonder why this is nice: There is the MiniTRX by OE3HKCm which is fully software-compatible to the mcHF but uses 7.4 V LiPo batteries...

I rather would let you do the changes than do it from our side right after integration.

mchf-eclipse/src/mchf_main.c
```diff
@@ -248,6 +248,7 @@ void TransceiverStateInit(void)
 
     ts.lcd_backlight_brightness = 0;			// = 0 full brightness
     ts.lcd_backlight_blanking = 0;				// MSB = 1 for auto-off of backlight, lower nybble holds time for auto-off in seconds
+    ts.low_power_shutdown = 0;                  // MSB = 1 for auto-off of backlight, lower nybble holds voltage x10 for shutdown threshold
```
db4ple: Please fix the comment.

mchf-eclipse/drivers/ui/ui_configuration.c
```diff
@@ -144,6 +144,7 @@ const ConfigEntryDescriptor ConfigEntryInfo[] =
     { ConfigEntry_UInt8, EEPROM_DETECTOR_COUPLING_COEFF_15M,&swrm.coupling_calc[COUPLING_15M],SWR_COUPLING_DEFAULT,SWR_COUPLING_MIN,SWR_COUPLING_MAX},
     { ConfigEntry_UInt8, EEPROM_DETECTOR_COUPLING_COEFF_6M,&swrm.coupling_calc[COUPLING_6M],SWR_COUPLING_DEFAULT,SWR_COUPLING_MIN,SWR_COUPLING_MAX},
     { ConfigEntry_UInt32_16, EEPROM_VOLTMETER_CALIBRATE,&ts.voltmeter_calibrate,POWER_VOLTMETER_CALIBRATE_DEFAULT,POWER_VOLTMETER_CALIBRATE_MIN,POWER_VOLTMETER_CALIBRATE_MAX},
+    { ConfigEntry_UInt8, EEPROM_LOW_POWER_SHUTDOWN,&ts.low_power_shutdown,0,0,255},
```
db4ple: You should set the default value here using the DEFINE.

phaethon:
I implemented your comments + some additional considerations. I don't think 0 is sensible starting value from usability perspective. Now I am using 6 bits + offset, which allows to change value from 6.5V to 12.7V. It is low enough for 7.4V Lipo. If needed we can move somewhat lower to 6.2-12.4.

I added logic to colour voltage based on the threshold value if set. Makes a simple visual notification if somebody looks into LCD. 

I did not want to create a separate configuration entry for enable/disable as too many settings decrease usability somewhat, but if others consider this as an improvement - it can be done of course.

db4ple:
Hi Eriks, 
thanks for taking my concerns into account. I also agree to "less is more" when it comes to configuration settings. I like the "automatic" voltage coloring when enabled. And now I am back at the discussion: If we would have the separate On/Off for the switch off, we could always use the coloring based on the set value. 
It would also allow us to let go of the offset altogether (7 bits and we simply limit the voltage range with min and max). I would see the "Under-Voltage Shutdown" in the standard menu, since this is something more often used (for portable use: on, for shack use: off), the voltage value itself in the configuration menu (since this will be changed rather seldom for a given machine and powersupply).

BTW, each configuration value may use upto 16 bits (even if we use an 8 bit data type). Storage is always 16bits.

Let's see what Andreas, DF8OE thinks.

df8oe:
Hi Eriks and Danilo,

I never have missed such a function because of if I am in portable use I am near to my rig and can see state of batteries in LCD. So I never missed an automatic function - but maybe under different usage profile it is a good solution. So I want to implement it. Because of storage always is in 16 Bit width we can use full range of an absolute minimum voltage (e.g. 2V) to maximum (e.g. 16V). Because of this firmware will be used in much mire rigs than mcHF in future we should be open for different hardware structures (wide range of voltage is recommended). I also would like to see two different settings (enabled/disabled) and voltage setting. Enabling/disabling should also be able to do by touching on voltage numbers on touchscreen. This is the fastest acess which is possible. Thank you for your contribution!

I do not have reviewed the code so I do not know if it is storing last settings and then powers down or it shuts down without saving to EEPROM. I am sure if this feature is released some OMs do want configuration of power off behaviour...

phaethon:
May be there are more portable specific settings? E.g. I set LCD blanking to off when operating from mains, and few seconds when from battery. There could be a setting: "Portable operation ON/OFF", which enables/disables  LCD blanking, low power shutdown, smth. else?
Or some settings could be saved separately for portable/non-portable operation. E.g. I use power 2W when operating from base (as I use additional amplifier), but 5W when portable. It could be saved automatically per mode.

db4ple:
Hi,

Shutdown is with "Saving Values" . 
-- off-topic on
BTW, after llooking at how we do this today I would rather change the  UiDriver_PowerDownCleanup(void) to UiDriver_PowerDownCleanup(bool saveConfiguration)  in the future. 
-- off-topic off
Regarding portable use: Yes, might be a good idea. However, this does not collide with having individual on/off switches in the configuration. If we expose these switches directly or indirectly will not change the configuration logic itself, only the code which controls the settings, and this gives us flexbility.

So let us go ahead here and finish this small addition.

df8oe:
I agree there are much more functions which should behave different if in home or in portable use. But because of everyone has different wishes it will be very difficult to find ONE setting which satisfies everybody. So many new configurations must be implemented - I do not think that is something we must think about at this time...

phaethon:
OK, as for this pull request, I will implement second configuration setting (ON/OFF) and the increased range.
If very high voltage can be selected for the maximum value, there is a risk that somebody will select 16V + auto-shutdown, and will not be able to even boot MCHF without a 16V power supply. For the start I will put limitation to 13.5V to lower the impact, and implement some safety check or override possibilities later.

As for the idea for separate settings for portable operations - I still believe into it, but I agree it should separated out for more consideration/discussion.

db4ple:
Hi Eriks,

great! Sanity check: The mcHF will always boot before it checks the voltage (and with F1-F3-F5 you can reset your configuration but this is of course a very extreme measure and should be avoided). I think the best way for now is to simply extend the "waiting time before switching off" and to add some message to the "switching off" screen. A minute wait will not hurt the battery much but allows to easily fix the setting if done wrong. If you want, I can do the "battery low shutdown message" part when I do my little refactoring of the PowerDown function.

phaethon:
@db4ple what is the preferred method to check if a minute has passed for this usecase?

db4ple:
Hi Eriks,

ts.sysclock is the global timer with 10ms resolution. 
The UiDriver_Timer struct and functions implement a simple timer abstraction. Not super precise but good enough. Maybe you want to use that like a watchdog (If above min voltage, rewind timer), if expired switch off.

phaethon:
Separated auto shutdown and threshold.
Implemented 60 sec delay from boot. The same variable and verification could be used in the future if delay after TX->RX will eventually be needed.

db4ple:
Hi Eriks, Andreas,

let us go ahead with the merge pull request. I will make a few changes (merge of the two I2C variables into one, use of two is kind of waste) but this is nothing serious. 

73
Danilo

db4ple:
Hi,
can we close this?

PR #927, opened 2017-07-01
Added 500 Hz tuning step

pa7n:
As most stations try to transmit on a round frequency number of 1.0 kHz or 0.5 kHz I added the 500 Hz frequency step. It's also a nice compromise in speed between 1 kHz and 100 Hz tuning.


df8oe:
I almost forget working in CW :/ Only working in SSB 500Hz stuning step is not so important as when working in CW!

DG9BFC:
just tried new version ... and i like the new tuning step of 500 hz ...

pa7n:
Nice you like it!
73, Erwin

PR #944, opened 2017-07-09
Splash Screen refactoring (Test required!)

db4ple:
The display is now initialized as early as possible.
A new "logging" function can be used during the boot phase to report
issues which in turn will be shown on the Splash screen.
At the end of the boot phase, the "logging" phase is finished
and normal operation is started.
In addition a "no display found" condition is reported (no green led during boot).
Since a screen is not necessary for operating the mcHF via CAT,
normal operation will be enabled despite the display being inaccessible.

PLEASE TEST IF SCREEN INIT WORKS on F4 SPI and PARALLEL.


DG9BFC:
seems to work
tested on mchf 0.4 with 2.8 display in parallel mode
greetz sigi

PR #967, opened 2017-08-01
Implementation of #966 (PTT via RTS)

db4ple:
RTS now can be used to control PTT.
Not tested, needs testing with RTS & DTR capable software (fldigi, WSJTX)
and with CAT software which uses the CAT approach (HRD, fldigi, WSJTX).

NEEDS TESTING, NOT TESTED ON LIVE SYSTEM!


ok1cdj:
Hi, will test it asap, when daily build will be available..

Tnx and 73 Ondra

PR #969, opened 2017-08-03
Corrected implementation to use the standard pattern

db4ple:


df8oe:
Ooouuuuuugh ;) That happens when I am soldering using left hand and programming using right hand at the same time....

PR #976, opened 2017-08-06
Fix issue with frequency bar not correct if CW Freq Offset == Shift

db4ple:
Now this mode is as usable as CW Freq Offset == RX.

What happens if a specific CW Freq Offset Mode is used, when switching From/to CW mode

Shift: shifts waterfall, displayed frequency doesn't change
RX:    doesn't shift waterfall, but displayed frequency changes

### DIFF
--- a/mchf-eclipse/drivers/ui/lcd/ui_spectrum.c
+++ b/mchf-eclipse/drivers/ui/lcd/ui_spectrum.c
@@ -1173,7 +1173,7 @@ static void UiSpectrum_FrequencyBarText()
         UiMenu_MapColors(ts.spectrum_freqscale_colour,NULL, &clr);
 
 
-        float32_t freq_calc = RadioManagement_GetRXDialFrequency()/TUNE_MULT;      // get current tune frequency in Hz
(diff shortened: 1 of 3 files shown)


df8oe:
Now everything works as expected and the two modes RX and SHIFT make sense:

If you use RX you can toggle from SSB to CW and if you have heard a CW signal in SSB it is same way heard when switched to CW (but frequency display jumps at height of sidetone). Absolutely normal: if you want to hear carrier it cannot be located at SSB carrier (because of this is at "0Hz"!!)
If you use SHIFT frequency display remains at toggling modes from SSB to CW - but now you have to change frequency (in height of sidetone frequency) if you want to hear a CW signal that formerly was heard in SSB.

Everything fine - mode TX can be removed (because there is no other useful mode than RX and SHIFT) and issue can be closed...

vy 73
Andreas

db4ple:
The funny thing is: there is virtually no specific code for the mode CW Freq Offset == "TX". It is "do CW" without CW specific handling for frequency display/tuning etc.

df8oe:
I think there has been planned something which never has been done :) So it is not astonishing that TX dos confusing things. I propose to eliminate this...

PR #987, opened 2017-08-14
Added an experimental rtty decoder (work for issue #9)

db4ple:
Enable in debug menu, 170 Hz shift, 2 stop bits, 45.45 Baud, center frequency
around 1000Hz (!). Can be used with the 500 Hz, Center 950Hz filter.

It is a very simple decoder, allows us to think how these
should be integrated.


DF5LI:
Cool! I've been waiting for a long time!!! 

73 Harri

db4ple:
Don't get too exited. I am not sure how it performs under real world conditions. Just tested with fldigi and about 5m distance between dummy load (transceiver with fldigi) and antenna (UHSDR OVI40).

df8oe:
It seems we are reaching the edge @ mcHF. If I turn RTTY on @ mcHF sound gets "rough" and mcHF is no longer accepting user inputs. The only way to leave this is removing power. It does not decode anything. I think STM32F4 does not have enough horse power to do the DSP jobs AND the RTTY decoding job...

If I turn on RTTY @ OVI40 it switches to RTTY and immediately starts to decode the noise :) I can change filters, frequencies and everything else. Added my antenna gives normal reception, activating my test TX running fldigi and working to a dummy load (which gives S9 at mcHF) results in 100% correct decoding. Very good start of work!

73
Andreas

DD4WH:
Hi Danilo, hi Andreas,
looks like a very nice start! However, I did not test yet. 
I am not convinced that we are reaching the edge, because as Danilo already stated in the code, the RTTY decoding runs at 48ksps! It would be easy to put that in the decimated audio path running at 12ksps (with a four times saving of processing time). AND:
Filtering is done twice: first the "normal" filters (hilbert + IIR lowpass/bandpass), then the two bandpass IIRs and a lowpass IIR in the RTTY-code. So lots of opportunities to optimize for the STM32F4. 
I will have a look, whether I can be of some help in making it run on the STM32F4, but have heavy time constraints right now . . .
73 de Frank
P.S.: 
CW decoding: https://sites.google.com/site/lofturj/cwreceive
PSK31 decoding: https://github.com/zenmetsu/radioModem
;-)

db4ple:
Yes, adding a decimation is a must, and maybe also switching to use the ARM functions. I think I can do this on my own as we have already decimation filters available. The RttyFilters can be recalculated easily using the mention URL with a filter calculator. I checked the page and I was able to get the values used in the code from the tool, so I am confident that I simply have to change the sample rate and will get the correct parameters for the change sample rate. I guess 8k sample rate is good enough and this will probably reduce performance requirements significantly (3 filters at 48ksps -> 3 filters at 8ksps + 1 decimation filter from 48ksps to 8ksps). I hope this reduces load in total a lot.

73
Danilo

db4ple:
Ok, I was out of sync, you already did what I proposed... cool.
TNX, Frank!

db4ple:
Hi Frank,
okay, you did half of it. We need to adjust the filter coefficients according to the decimation rate, otherwise it is not working. I can do the rest as explained above. But not now...

DD4WH:
Hi Danilo,
no, I did not do it correctly, because I did not change the filter coefficients ;-).
So, go ahead, and change the filter coefficients with the filter tool!
Two comments:
1.) 8ksps is not easily feasible if we use the ARM functions, because decimation factor has to be a multiple of the blocksize! So decimation-by-4 is fine, decimation-by-6 is not easy. Why not use decimation-by-8? Would mean 48/8 = 6ksps, that means max. frequency usable is 3k, should be enough! Or just use the decimation already done in the audio path (by 4).
2.) the lowpass filter with corner frequency 50Hz using an IIR filter will not work with single floats! You need more precision (double) with such low corner frequencies. So we will have to use a FIR filter OR use double precision ONLY for that specific filter.

73 de Frank

PR #990, opened 2017-08-15
experimental RTTY: added shift & freq for DWD weather reception 

DD4WH:
does not work (yet!?):
To try, uncomment "#define DWD_STANDARD" in audio_driver.c line 1776
- I changed frequencies to 915Hz and 1365Hz = shift of 450Hz
- I changed Baud rate to 50Hz
However, does not work with DWD reception, maybe you have to change lines 1949 ff., because no. of samples per bit changed with decimation/sample rate??
try 
LSB 4584.035kHz
USB 10999.765kHz
LSB 7647.035kHz
USB 11037.965kHz
USB 14.466.265kHz

### DIFF
--- a/mchf-eclipse/drivers/audio/audio_driver.c
+++ b/mchf-eclipse/drivers/audio/audio_driver.c
@@ -1761,6 +1761,7 @@ float32_t hang_backmult;
 float32_t onemhang_backmult;
 float32_t hang_decay_mult;
 
+//#define DWD_STANDARD
 
 // RTTY Experiment based on code from the DSP Tutorial at http://dp.nonoo.hu/projects/ham-dsp-tutorial/18-rtty-decoder-using-iir-filters/
 // Used with permission from Norbert Varga, HA2NON under GPLv3 license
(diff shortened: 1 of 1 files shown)


db4ple:
We need 2 stop bits, probably DWD uses 1.5 which we will not decode correctly it all (Tried with fldigi, we reject this).
I would ask to get permission to improve the handling of different modes (no recompile necessary). Ok?

DD4WH:
OK, so that´s a pity ;-). Have not understood the concept of half bits, can you recommend something to read about that? Is it possible to easily extend the code from 2 to 1.5 stop bits, or is that a completely different scenario?
Yes, OK, go ahead!

DD4WH:
I am thinking about making the filters more flexible, so we can potentially cover many more digital modes without having to store specific coefficients for each mode. But I will first make some kind of a plan and will not touch the code.

db4ple:
Well 1.5 stop bits is just that, the last stop bit is only half as long as the normal bit. Or the only stop bit is exactly 1.5 bit times long. Same thing.

I just separated filter config from code, makes playing around much easier.  Will test and commit.

PR #993, opened 2017-08-15
Refactored the RttyDecoder to use filter configurations

db4ple:
- New filters can be easily added and changed during runtime (code for changing fully not implemented)
- Only 2 stopbits implemented
- seemingly much higher performance than hardcoded filters, probably due to
  keeping the filter parameters in RAM.

Decode @ 12k works (@ 48k also).

### DIFF
--- a/mchf-eclipse/drivers/audio/audio_driver.c
+++ b/mchf-eclipse/drivers/audio/audio_driver.c
@@ -499,15 +499,6 @@ void audio_driver_config_nco(void)
 //
  */
 
-//
-//*----------------------------------------------------------------------------
-//* Function Name       : audio_driver_init
-//* Object              :
-//* Object              :
-//* Input Parameters    :
-//* Output Parameters   :
-//* Functions called    :
-//*----------------------------------------------------------------------------
 void AudioDriver_Init(void)
 {
     const uint32_t word_size = WORD_SIZE_16;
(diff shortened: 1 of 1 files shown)


DD4WH:
Hi Danilo, very nice, thanks a lot for your contribution. It is really fun to play around with that.
"higher performance": could that be caused by a swap of mark and space? 
############ 
-    float32_t line1 = RttyDecoder_bandPassFreq0(sample);
+    float32_t line0 = RttyDecoder_bandPassFreq(sample, rttyDecoderData.BP0_coeffs, &rttyDecoderData.BP0);
-    float32_t line0 = RttyDecoder_bandPassFreq1(sample);
+    float32_t line1 = RttyDecoder_bandPassFreq(sample, rttyDecoderData.BP1_coeffs, &rttyDecoderData.BP1);
############
73 de Frank

db4ple:
No, I did not swap mark and space, I just corrected a quick-and-dirty hack. The original decoder was for mark -> high and space -> low but in LSB (FSK) we need mark -> low and space -> high. Initially I did switch the line assignments to do this, now I cleaned up and named the more stuff more appropriately. Please rebase, BP0 and BP1 are gone now and are named bpfMark and bpfSpace.
If this is better understandable, we'll see. But the commit also supports 1 / 1.5 stop bits, so you may want this to check out the DWD reception.

db4ple:
I tested the different implementations side by side:
F7 original BP/LP implementation 10% additional load @48ksps
F7 new BP/LP implementation 5% additional load  @48ksps (i.e. 50% less load compared to original)
F7  new BP/LP implementation 2% additional load  @12ksps (i.e. 80% less load compared to original)

PR #1001, opened 2017-08-17
Entering characters using paddle in iambic mode

phaethon:
This is currently not meant for merging, but for feedback and discussion. It is part of my overall plan to implement PSK functionality. ui_driver.c is just a hack to see the code working, which is not meant for production code.
The code detects characters entered using paddle in iambic mode. This can be used anywhere where text input is required digital text modes (RTTY, PSK, etc.), memory editing, showing feedback of entered text in CW mode, etc. At present only alphabetic characters are decoded, but adding other symbols and prosigns is easy.
@db4ple I see you have committed RTTY decoder.


phaethon:
See #1002 for wider context.

db4ple:
Hi,

very nice idea. Seems I have to learn morse code now. Good that we have low WPM settings ...

Anyway, I think we should follow-up that idea. Extremely ham-style input. 

However, I would ask you to use rebase instead of merge. This prevents use from resolving merge issues. I know, you said, it is not for merging yet. But I guess there is a big demand coming up. I can see it being used immediately to monitor your CW input while you are TXing. Extremely simple to do. We can just use the very same function we use in RTTY and FreeDV to run text on the screen (you just send a character to it, it will display it and handle line feed, scrolling etc on its own).

The rebase procedure is described here: https://github.com/df8oe/UHSDR/blob/active-devel/CONTRIBUTING.md

73
Danilo

db4ple:
BTW, the functions name is UiDriver_TextMsgPutChar(charResult).
The output is regularly displayed on screen (if there is new text), nothing more to do for you to get unlimited CW output on screen.

df8oe:
Merging conflict is not as hard as that I cannot merge it (and I WANT TO DO THIS!!) But i agree that it is a much better idea first to rebase, then use the new text output method and commit again. Absolutely nice. I like it!

phaethon:
See the updated code. Rebased the code. Removed text buffer, calling UiDriver_TextMsgPutChar instead. Added digits and symbols to the table of characters (no prosigns yet). 
I must admit, I have not tested this version of code on a device yet, but it shouldn't be dangerous.
One more note - it works only with paddle in iambic mode. Making it work in straight key mode would be more tricky and would require somebody else to test it, as I am using only iambic mode myself. On the other hand it is, of course, doable.

df8oe:
*Very Very* nice. But Ultimatic mode is damaged now... If you press & hold dith and then press dah it hangs. Normally it must produce dahs until you release dah paddle...

df8oe:
I just stated error is in line 699 of cw_gen.c. If I use there old value Ultimatic mode works again. But I have long time not done CW so I must say that there is another error which must have to with CW weighting. I can use Ultimatic for a handful of signs and then it hangs again. This has nothing to do with this commit - I think.... I will open an issue.

PR #1005, opened 2017-08-17
Small bugs in the Ultimatic

phaethon:
Sorry for the bug with ultimatic. There was an additional small part of redundant code, which this PR fixes.
Additionally, I have an issue with space. If the txrx break time is shorter than the length of interword pause, then the code adding space is never executed. But for this I have not thought of a good fix yet.

### DIFF
--- a/mchf-eclipse/drivers/audio/cw/cw_gen.c
+++ b/mchf-eclipse/drivers/audio/cw/cw_gen.c
@@ -78,6 +78,7 @@ typedef struct PaddleState
     // Timers
     ulong   key_timer;
     ulong   break_timer;
+    ulong   space_timer;
 
     // Key clicks smoothing table current ptr
     ulong   sm_tbl_ptr;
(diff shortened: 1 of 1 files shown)


db4ple:
Hi Erik, good work, but we have to talk a bit. There are a few code style guide lines we try to enforce, some are more of cosmetic nature (where are parentheses are placed), some are let's say more based on experience when it comes to write robust. I hope you don't mind to follow these rules (once I documented them properly in the Wiki). I think, it should be said, that some of the rules may force you to things a little bit different here in this project but we want to get as consistent and robust code as possible.  

Otherwise the coding style police will put hand on your code and change it for you. But this is not a nice job, no one wants it.

phaethon:
Sure, it is ok for me to follow the rules. Send me or give a link to read it.
On my part, I would like to more on your plans with RTTY for clarity, as we both recognized some parts are potentially overlapping.

df8oe:
I want to test this and merge it. Ultimatic always was problematic and maybe is not perfect at all but it is the mode I am working with :) It is not a problem to correct spaces and code to follow the programming guiding lines later...

phaethon:
That's the issue with cw paddle code development - I can meaningfully test only the mode I am using :)

PR #1006, opened 2017-08-18
Added some code style info

db4ple:
See this as input / first draft.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,7 +4,7 @@ This document describes some points about the contribution process for mcHF firm
 
 The maintainer of this GitHub is DF8OE.
 
-The project is free for non-commercial use and is being developed within an open community. Maintainer merges pull-requests, fixes critical bugs and manages releases. Participation is appreciated.
+The project is licensed under GPLv3 and is being developed within an open community. Maintainer merges pull-requests, fixes critical bugs and manages releases. Participation is appreciated.
 
 ### Getting Started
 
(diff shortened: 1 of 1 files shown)


phaethon:
Can you explain benefit of single return statement? In my recent code I used return to break out of loop (and return value). The alternative would result in redundant state variable, which would need to be used and redundant code to check if the variable was assigned.

db4ple:
More than happy, although your function is not on the complex side:

Your code:
```C
void CwGen_AddChar(ulong c) {
    for (int i = 0; i<CW_CHAR_CODES; i++) {
        if (cw_char_codes[i] == c) {
            UiDriver_TextMsgPutChar(cw_char_chars[i]);
            return;
        }
    }
    UiDriver_TextMsgPutChar('*');
}
```

Code with single return:
```C
void CwGen_AddChar(ulong c) {
    char val = '*'
    for (int i = 0; i<CW_CHAR_CODES; i++) 
    {
        if (cw_char_codes[i] == c)
       {
            val = cw_char_chars[i];
        }
    }
    UiDriver_TextMsgPutChar(val);
}
```

Advantages: 
- Single place for processing of calculated value (Function call). Easier to maintain. Less code generated by compiler in most cases. 
- No need to look for other places where the function is left. This is especially problematic for longer functions with switch case etc.  
- Compilers are extremely good at optimising code given the opportunity. Especially variable assignments.
I can assure you that "val" is never a true variable in memory, this will live in a register of the processor. So the resulting executed code is most likely very similar (but not identical as we have only a single function call but even this could have been done by the optimizer. You can try to compile both. And watch the size of the executable.

BTW, this rule comes from MISRA, a C coding standard for embedded systems development. Not all MISRA rules should be followed but I think the single return rule makes sense and is in most cases beneficial in most cases (of course there are exceptions):

https://spin.atomicobject.com/2011/07/26/in-defence-of-misra/

And for another opinion:

https://www.misra.org.uk/forum/viewtopic.php?t=928

I can tell you, converting the code in the mcHF firmware to use single returns was helpful for me to understand certain problems in the firmware. 

Think of real-time debugging. If I want to know what is the value of a local or global variable at the end of the processing of a function, with single return -> I place the breakpoint at this return. With multiple returns I have to do this in many places. And we have only hardware 4 breakpoints in the STM32F4 (since we run our code in flash).  And you can find more benefits. Multiple returns are a little bit like "goto", which is also something I see rarely a use for. I spent hours on debugging wondering why I never reached certain pieces of code (because some hundred line above code decided to jump out of a nested loop straight away). It may be me to be not clever enough to handle this. 
I fully agree to the statement in one of the mentioned posts: If it doesn't work with a single return, consider to refactor.

phaethon:
Simple cases are easier to discuss :)
You needed to introduce additional variable. In this case it could go into register, but there is a limited number of registers, using one more may mean some other optimization does not happen. Only analyzing limited part of compiled code, not always reflects actual impact in a large code. Also, having tried to get into functional programming - any additional variable is 'a sin' by default :) and there is some reasoning for that, too.
For loop needs to check the full array in your example. Otherwise it takes on average 2x the time than the original code. `break` should be added within conditional block to end the loop sooner.
It would be more complicated in a nested loop case as you would likely need additional bool variable to break from double loop + conditional.
So, as you made the code in the example 2x slower, I wouldn't agree that overall it is improvement by default :)
Let's say that for this specific style guide rule, I will always give a second thought and at least consider refactoring if I see a need of several return statements. For the rest of the style guide - no objections.

PR #1015, opened 2017-08-24
RTTY: added ATC = automatic threshold correction

DD4WH:
ATC should lead to more robust decoding of RTTY under fading conditions. See this page for more information: http://www.w7ay.net/site/Technical/ATC/
I have simply taken code bits from FlDigi source code and adapted them to UHSDR code.
For comparison, we now temporarily have two digital modes: "RTTY" and "RTTY ATC", so we can compare the performance of RTTY without ATC ("RTTY") and RTTY with ATC ("RTTYATC").

Tested:
DWD weather report decoding in RT-U. No significant differences with 1k4 BPF. But try to switch to  500Hz/950Hz filter (which leads to space being OUT of the filter passband): then RTTYATC has a huge advantage over RTTY. The latter is not able to decode anything when only mark is present, whereas RTTYATC even pulls out some content when only mark is present . . .
Not tested:
Ham Radio RTTY

### DIFF
--- a/mchf-eclipse/.settings/language.settings.xml
+++ b/mchf-eclipse/.settings/language.settings.xml
@@ -5,7 +5,7 @@
 			<provider copy-of="extension" id="org.eclipse.cdt.ui.UserLanguageSettingsProvider"/>
 			<provider-reference id="org.eclipse.cdt.core.ReferencedProjectsLanguageSettingsProvider" ref="shared-provider"/>
 			<provider-reference id="org.eclipse.cdt.managedbuilder.core.MBSLanguageSettingsProvider" ref="shared-provider"/>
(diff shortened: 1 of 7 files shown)


db4ple:
Cool stuff, but you could have use a simple debug on/off menu entry to turn on and off ATC. No need to introduce a new mode. If I may, would like to do this change for you or you do this. It might be a little more convenient to switch by touch with the two modes but on the long run, we will not use two modes for RTTY but rather a mode specific setting (or use the ATC all the time).

DD4WH:
Hi Danilo, yes, I can do this change. But I would wait until users have tested and we have decided whether RTTYATC is really worth the effort or if we stick with RTTY. I surely did not want to introduce a new mode, it was just one way for me to have the two modes at hand for comparison temporarily.
In a few days, when we have more test results, I/we will throw out one of the "modes".
BTW, do you have any plans or already begun to build a CW decoder?

db4ple:
a) no, not yet CW, go ahead if you like. 
b) The new mode is a problem for a change I want to implement: Storing the current digital mode. With new modes which is there only for a brief period, the mode numbering is screwed up. Since 3 (which is now RTTYATC) may be later PSK and so forth. It is simply not a good idea to "randomly" introduce new modes. You can of course argue, there is no other mode yet. 
The "variable" approach would also have much less impact on existing code, look at all the places you had to handle the RTTYATC mode.

DD4WH:
a) well, you prepared everything so well, maybe I will try something, but no promise ;-)
b) I am convinced, will change that this evening as proposed!

DD4WH:
b) done!

PR #1017, opened 2017-08-24
put RTTY ATC into Debug menu and erase new mode RTTYATC

DD4WH:
ATC for RTTY is now available for testing in the debug menu. No need to have it as a new "digital mode", as this could potentially spoil up ongoing advance in saving digital mode setting in EEPROM.


db4ple:
Thank you! 🥇

Issue #1002, opened 2017-08-17
Implementing PSK mode


phaethon:
I have started the work on implementing PSK mode. The objective is to be able to fully communicate in PSK without using a computer. There is a multitude of design issues.
1. Selecting the PSK mode
Should it be a separate PSK mode or a submode of existing Digital mode? If the latter, how to select it?
There is BPSK, QPSK, 8PSK with many variations (e.g. error correction option) and different speeds (BPSK31, BPSK63, etc.). How to select specific PSK submode? For the start I prioritize implementation of BPSK, and the speed to be selectable (and displayed) the same way as WPM in CW. Implementing other PSK modes should be more a UI issue as the rest should be easy.
2. Displaying decoded (and transmitted) text
Should it be separate display mode with text window, or an overlay over existing waterfall and scope? Showing received text and transmitted text in separate windows or in one with different colours? What variations/configuration options would be needed to cover different preferences (e.g. font size)?
3. Entering the text
I have already created a prototype paddle iambic mode input decoder. Probably, straight mode decoder would be needed for some. USB keyboard might be very convenient, but I understand that constantly polling keyboard could be a performance issue.
Also, it should be possible to switch between tx and rx modes. Likely using one of the physical buttons.
There should be a buffer of text, which is displayed and sent once switching to tx mode, and some very simple editing possibilities should exist (cancelling all text, deleting last character).
4. Memory keyer buttons
Using PSK without possibility to send a predefined block of text would not be convenient and wouldn't replace computer usage. One idea is to use long press power button to switch 5 buttons into 4 memory buttons (sending the recorded text when pushed) + 1 tx/rx button. Long press on power button again switches back to normal arrangement with Menu, Meter, etc.
Memory keyer buttons could be interesting for CW mode as well.
Ideally, it would be possible to select callname from conversation (probably using touchscreen), and simple macros could be supported. This is more for a later phase, and not for the start.
5. Simple logging
This is optional, but as we have frequency, time (if using RTC), band, mode, call, we can keep a simple list of QSOs . Not doing anything now, but returning to this idea in the future.

Feedback welcome.

db4ple:
There are already provisions in the digital mode handling code for multiple digital modes. I am about to integrate RTTTY into that framework. That means I basically have the same questions you have.  

1. I think we should, like in CW use some of the encoders to select some properties of the digital mode.
[...]

PR #1026, opened 2017-08-31
First commit for keyer mode implementation

phaethon:
This is the start of the keyer mode implementation as discussed in #1002 
Long press on M1 (if not in menu mode) switches to-from keyer mode. In keyer mode F buttons are remapped:
F1-F3: keyer buttons BTN1-BTN3 (short press plays, long press records)
F4: deletes last character from buffer on short press, deletes buffer contents on long press
F5: switches tx/rx mode in buffered mode, on long press - switches on/off buffered tx on 
This PR implements only mode switching and button labels. 
@db4ple , please, consider using for RTTY.


df8oe:
@phaeton: Would you please be so kind and add operating instructions to our WIKI?

db4ple:
Hi Eriks,
thanks. Please try not to work in ui_driver.c in existing functions for the moment I have made significant changes to the way how keyboard and touchscreen events are processed. I will need to integrate your changes carefully as the functions you changed no longer exist in their previous form (as you have used them).
On the other hand, implementing alternative keyboard functions will be much easier.

BTW, please remember, we strongly ask users to use regular rebase throughout the development of their changes instead of merge of changes at the end of the own development.

phaethon:
@df8oe I will add instructions to Wiki as soon as I have the first implementation of the actual buttons.
@db4ple The only part of existing button functions that I plan to touch in short term are the placeholders with //NOT IMPLEMENTED. That was half of the reason for the commit - to make it easier to merge. I will try to switch to rebasing from now on.

AsbjornPettersen:
I see that 2 variables have been added in TransceiverState:

bool keyer_mode; // disable/enable keyer mode for F1-F5 buttons
bool buffered_tx; // disable/enable buffered sending for CW and digital modes

There are hundred of items in TransceiverState and if buffered_tx only shall be used for "keyer_mode" a struct should be made:

struct f_keyermode
{
	bool keyer_mode; // disable/enable keyer mode for F1-F5 buttons
   	bool buffered_tx; // disable/enable buffered sending for CW and digital modes
};
struct f_keyermode fkeyer;

When keyer_mode is set to false, the buffered_tx also should be set to false !?

phaethon:
No, these are not related. There could be buffered tx without keyer, and vice versa. 
Buffered tx is related to digital text mode sending. We can merge it in struct with other digital mode variables.
Keyer mode is related to menu mode as it also remaps f1-f5. Actually menu mode could be renamed into eg interface_mode with values 0,1,2 dropping keyer mode entirely, but I am not sure if we want this. Although, it could look nicer in F button processing with a switch instead of cascading ifs.

db4ple:
Please review the changes in my pull request #1027. I prepared an "empty" set of keyactions_keyer. I think you will quickly get the idea. A KEYACTION_NOP  is a NOP but does mark this key press as being processed, KEYACTION_pass leaves the key as being unprocessed, i.e. in another list which is being processed later, this key may still trigger an action.

What is not yet handled well is the update of the respective Fx key labels, but I am not too concerned about this right now. 

BTW, please respect the code style as used in the project so that we do not have mixed styles of parenthese etc. We are all asked use the same formatting, even it is not our personal preference (it would not be mine, to be honest). Acceptance of compromises is a vital part of successfull collaboration.

PR #1033, opened 2017-09-04
Refactorings, small functional changes

db4ple:
- Aligned naming of all exported uhsdr_board.c functions
- Reduced use of non_os_delay(), some locations removed, some replaced by HAL_Delay
- Refactored the Si570 interface to be slightly simpler

PLEASE TEST IF TX ESPECIALLY WITH MICROPHONE IS STILL OKAY!


df8oe:
There is an issue (#1035) at your commit or phaetons #1034. Please take a look.

PR #1034, opened 2017-09-04
Fixing cw entry of space, now calling cw routines not only in CW/TX mode #1002

phaethon:
Fixed space issue. Now, a keyer memory can be saved with spaces. Also, decoding spaces correctly when in cw/tx.
Important change is that cw processing is called also when in cw/rx mode, or when cw_text_entry and non-cw.


df8oe:
There is an issue (#1035) at your commit or Danilos #1033. Please take a look.

PR #1036, opened 2017-09-06
CW decoder first attempt - not functional

DD4WH:
not yet working - will work on this in the next days / weeks


db4ple:
Hi,
let us keep the Cw Decoder as much as possible outside of the audio_driver.c (like with rtty ).
I would do this for you as a next step after the pull request has been accepted (or you do it before).

73
Danilo

DD4WH:
Hi Danilo, fully agreed. I tried that in my first attempts, but did not succeed with all these includes etc.
Could you do that for me, please? Would be very much appreciated :-).
Also, I plan to use the character codes in the CW part (and not another redundant character array), but that´s for later, first I would like to try to get the whole thing functional, which it isnt at the moment.
73 Frank

PR #1040, opened 2017-09-08
CW decoder experiment

DD4WH:
Many things in the code still not clear to me. Some decoding is being done, but I have no CW signals at the moment to test.
1 character is always being displayed in the second line, I do not know whether this is a race condition, a code error or a property of the UiDriver_TextMsgPutChar - code !?

### DIFF
--- a/mchf-eclipse/drivers/audio/cw/cw_decoder.c
+++ b/mchf-eclipse/drivers/audio/cw/cw_decoder.c
@@ -36,7 +36,7 @@ void CwDecode_FilterInit()
 #define CW_DECODER_AGC		0
 #define CW_TIMEOUT			3  // Time, in seconds, to trigger display of last Character received
 // and a New Line in the USB Serial Monitor.
-#define ONE_SECOND			375 // 375 * 2.67ms equals one second
+#define ONE_SECOND			(12000 / CW_DECODE_BLOCK_SIZE) // sample rate / decimation rate / block size
 #define CW_AGC_ATTACK      	0.95  // Audio automatic gain control (AGC) attack, audio vol reduce per cycle.
 #define CW_AGC_DECAY      	1.005  // Audio AGC decay, audio volume increase per cycle.
 // AGC attempts to cap the max signal level at the Fpeak frequency to 40
@@ -211,12 +211,12 @@ static void CW_Decode_exe(void)
 	if (siglevel >= thresh)
 	{
 		cw_state = TRUE;
-//        Board_RedLed(LED_STATE_ON);
+        Board_RedLed(LED_STATE_ON);
 	}
 	else
 	{
 		cw_state = FALSE;
-//        Board_RedLed(LED_STATE_OFF);
+        Board_RedLed(LED_STATE_OFF);
 	}
 #endif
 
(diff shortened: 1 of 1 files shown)


db4ple:
Hi,
it should not be UiDiver_TextMsgPutChar -> the length is limited to fit entirely on a single line.

DD4WH:
I dont know what you mean exactly: as far as I know, the characters are sent one by one, so I expected  UiDiver_TextMsgPutChar to cope with all that!?

db4ple:
I meant the character on the second line should not come from the PutChar call.

DD4WH:
oh, then I will scan the code again, because I meant to only use UiDiver_TextMsgPutChar for printing. Will dig deeper into that.

db4ple:
You don't have to scan, there is nothing in it. I did that already.

db4ple:
Can you send a screenshot of the problem with the character on the second line?

DD4WH:
sent you a PM, that was easiest.
The characters in the upper row scroll like they should, the character/s in the second row stay where they are. They seem to appear only the first time the character row is filled, this happens really fast.

DD4WH:
will be busy with other things now, I can do more tomorrow :-)

db4ple:
Hi, I checked, unfortunately, my display does not have the problem. So I cannot reproduce. However, what I found is that I had a total lockup of the TRX. Thanks to the ST-Link it was easy to find out.

Line 1144 in cw_decoder.c:

while((!DataRecognitionFunc()) && (ts.sysclock < freezetime + 2));

This will not work as expected. You are calling the CW decoder in the Audio-Interrupt. And this kind of loop will not work: the Audio-Interrupt itself is responsible for increasing ts.sysclock. So if you wait inside the Audio-Interrupt for it to change: Will not happen.  Generally speaking: Any form of busy waiting should not happen inside the interrupts. In fact, it should never be used unless there are very good reasons as this stops the other functions of the TRX to get their time. ts.sysclock increments in 10ms steps, this is very long.  
Code using such an approach is not directly usable for us.

db4ple:
A quick scan through the code shows that the error correction function is the only one doing such "unwanted" waiting. So the rewrite has to enable us to recognize that we are in "error correction" mode and should be able to continue where it left off last time. There are three locations of such a loop in the error correction function. This is not going to be trivial.

DD4WH:
OK. Thanks a lot for identifying! I inserted that ts.sysclock waiter assuming that it would be incremented by another interrupt and not the same interrupt it was implemented in. So I will go deeper into that and have a close look at the error correction function and what could potentially hang there.
Strange things seem to happen with my particular display hardware . . . it makes problems with the noise blanker and also with this text display, I wonder what´s going on there . . . !?

db4ple:
I think I found a fairly simple solution to the problem. We don't need the timer, we simply can check if the function is stuck in an potentially endless loop.

I can commit a pull request containing the required changes. It is not much.

DD4WH:
That would be great, if you could include that endless loop check!
No more chance to test the decoder and parameter settings on the bands tonight, HF conditions are extremely bad at the moment, Aurora is there and all bands seem pretty closed! ;-)
Good night!

PR #1083, opened 2017-09-18
Implemented  "Keep Signal Option" for CW Shift Offset Mode

db4ple:
What it does: Best of both worlds (in a way)
Like in RX or TX, if you are tuned to a signal around the CW sidetone frequency in SSB
you will stay tuned to this signal if going into CW mode.
I.e. if your sidetone is 700 Hz and you are tuned to 7.000.000 Hz and you hear
a CW signal with about 700 Hz, simple switch to CW. You will still hear the signal
and your frequency display will show 7.000.700 Hz, since this will be your TX
frequency.

Unlike CW RX or TX offset modes it works even if you are switching sidebands, e.g. from
SSB LSB to CW-U (USB). Also if you switch between CW-U and CW-L you'll stay
tuned to the signal (as it was in CW Shift Mode). If you switch back to SSB,
you will still be tuned to that signal.

How to operate: In CW menu, set CW Offset to "Shift". In Debug menu, switch on "CW Shift Keeps Signal".
Tune to a CW signal around the sidetone frequency, and switch to CW.

@s53dz: Give this a try and tell what you think.

### DIFF
--- a/mchf-eclipse/drivers/ui/menu/ui_menu.c
+++ b/mchf-eclipse/drivers/ui/menu/ui_menu.c
@@ -1747,24 +1747,24 @@ void UiMenu_UpdateItem(uint16_t select, uint16_t mode, int pos, int var, char* o
         const cw_mode_map_entry_t* curr_mode = RadioManagement_CWConfigValueToModeEntry(ts.cw_offset_mode);
         temp_var_u8 = curr_mode->sideband_mode;
 
-                var_change = UiDriverMenuItemChangeUInt8(var, mode, &temp_var_u8,
-                        0,
-                        2,
-                        2,
-                        1
-                );
+        var_change = UiDriverMenuItemChangeUInt8(var, mode, &temp_var_u8,
+                0,
+                2,
+                2,
+                1
+        );
 
-                if(var_change)      // update parameters if changed
-                {
-                    cw_mode_map_entry_t new_mode;
-                    new_mode.dial_mode = curr_mode->dial_mode;
-                    new_mode.sideband_mode = temp_var_u8;
-                    ts.cw_offset_mode = RadioManagement_CWModeEntryToConfigValue(&new_mode);
+        if(var_change)      // update parameters if changed
+        {
+            cw_mode_map_entry_t new_mode;
+            new_mode.dial_mode = curr_mode->dial_mode;
+            new_mode.sideband_mode = temp_var_u8;
+            ts.cw_offset_mode = RadioManagement_CWModeEntryToConfigValue(&new_mode);
 
-                    ts.cw_lsb = RadioManagement_CalculateCWSidebandMode();
-                    UiDriver_DisplayDemodMode();
-                    UiDriver_FrequencyUpdateLOandDisplay(true); // update frequency display and local oscillator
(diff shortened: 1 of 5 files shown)


s53dz:
Hi Danilo,

TNX for providing that.
Just a quick preliminary report:
with CW-L and LSB works but with USB and CW-U does not.

**EDIT:**
As I see it with USB it should be switched to -750 Hz? Guessing. 

73 Bojan

[CW_display1.pdf](https://github.com/df8oe/UHSDR/files/1311690/CW_display1.pdf)

db4ple:
Thanks for testing. It does not seem to work if you do the following:

LSB -> CW-L -> CW-U -> USB  or USB -> CW-U -> CW-L -> LSB

It currently only works as if you leave to the same sideband-mode you came from because we restore the exact same frequency.  Since I have LSB/USB autoselect for SSB on, CW brings me back to the same SSB sideband. In your case you have to go from USB to LSB and voila, you have your signal back.

I don't think we should change that (although we could). It is difficult to predict which sideband will be following CW since we may have several non-sideband modes ahead of us. 

What we should do, is to return to the last selected sideband mode in SSB instead of going
LSB -> Long Press -> USB -> Tune CW -> Short-Press -> CW-L -> LSB -> then we loose the signal until we do a long press in LSB to go to USB.

s53dz:
Ok, my bad.
I thought you were following Andreas's idea of maintaining the signal when changing CW-L <--> LSB bellow 10 MHz and CW-U <--> USB when up.
Now when the tune helper works in both side-bands it is so and so with less importance since we can return to the CW_LSB/USB=auto and LSB/USB mode change=auto. Normally we do not change this setting when operating although there is no other barrier than the historic pre-definition.

As I said, for me it is really important to see displayed frequency which is in line with spectrum and that that is also an actual RX and TX. Now you managed also to maintain the signal when bellow 10 MHz. It should be as easy as possible.

73 Bojan

db4ple:
The "Keep Signal" approach works currently works best if used with the a single sideband mode for SSB listening (e.g. by using the LSB/USB autoselect). The sideband approach used in CW is irrelevant, keep signal works with all of them, you may use always CW-U or always CW-L or use autoselect CW-U/L although this is not necessary.   

I am absolutely with you, when you say, it should be as easy as possible. 

I am trying to figure out if we can come to a what I call "Unified CW Sffset Mode" (i.e. a single way to handle switching to and from CW acceptable for all CW users) so that we can remove the three choices (RX,TX and SHIFT) which cause a lot of confusion and discussion (and make the code difficult to understand).

s53dz:
Correcting my report again,
it works well in USB and in LSB (tuned signal stays in when changing mode), except immediately after band switching it changes frequency (probably for 2xSidetone) and it does not work!? After re-tuning to the frequency it works again either in USB as in LSB. Strange.

Actually, each time you **switch between bands** and the **mode is not the same**, you are running off the frequency again and again.

73 Bojan

db4ple:
I should have said that switching bands is not supported right now (since this will lead to unexpected frequency changes). This not a limitation of the approach as such just a lack of complete implementation since I wanted to run the idea through some users.

PR #1088, opened 2017-09-21
Replaced a number of double float functions with their single precision float counterparts

db4ple:
I did this only in places where only float32_t seems to be involved.
Should not change a thing. Reason for change: code size reduction (a few kBytes).
Speed improvements (marginal at best).

I used fw-mchf.map to identify which double functions are still in use.


s53dz:
I suppose this (please, see the attachment) is not connected to your last optimization, however I get these warnings when building (STM32F4) using gcc...ver.6.3.1 and none when using ver.5.4.1.

73 Bojan

build2

df8oe:
This messages are present since we switched to HAL. They are related to a library - no code produced by us :)

You can ignore them. This is not lucky but actually there is no other possibility.

s53dz:
Andreas,
thanks for your explanation. So, it is better knowing that I am not alone with the pain of getting many warnings.

73 Bojan

PR #1094, opened 2017-10-01
CW decoder fine tuning and clean ups

DD4WH:
- experimental CW decoder AGC --> enable in CW menu
- CW tune helper now enabled on default (and possible to choose in AM / SAM)
- renamed functions and variables
- finetuned settings for CW decoder
CW decoder seems to work reasonably well now with manual threshold adjustment
CW decoder with CW decoder AGC switched ON in CW menu does not work as well as I want it too (yet;-)

### DIFF
--- a/mchf-eclipse/.settings/language.settings.xml
+++ b/mchf-eclipse/.settings/language.settings.xml
@@ -5,7 +5,7 @@
 			<provider copy-of="extension" id="org.eclipse.cdt.ui.UserLanguageSettingsProvider"/>
 			<provider-reference id="org.eclipse.cdt.core.ReferencedProjectsLanguageSettingsProvider" ref="shared-provider"/>
 			<provider-reference id="org.eclipse.cdt.managedbuilder.core.MBSLanguageSettingsProvider" ref="shared-provider"/>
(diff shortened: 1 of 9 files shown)


phaethon:
@DD4WH Do you plan to look into possible merging of CharacterIdFunc and encoding method used in cw_gen.c or shall I look into it? 
Also, it would be nice to apply cw decoding to straight key keyer decoding.

DD4WH:
Hi Eriks,
I have it on my ToDo list, but I fear I have not understood your code/encoding method good enough to be able to do it ;-).
So, if you would be willing to do it, please go ahead! That would be really great!
I will be working on SNAP (i.e. a one button tune helper for CW) in the next days and will not touch the CW decoder code, except for very minor things.  
73 Frank

db4ple:
@phaethon: Hi Eriks,
I looked into this: CharacterIdFunc: And I think we should do it. Feel free to go ahead.

A small thing to take care of is to provide the right code back in case of "no decode". 
My main concern is the fact that the CW decoder code has the ability to store upto 40 symbols (dit/dash). With your approach we currently can store only 8 symbols in 16bits (with 32bit 16 and with 64bit 32). I haven't yet figured out, if there can ever be a need for storing  40 symbols. I would say no since we can store only 40 times of low and high which can result in only 20 symbols. But I am not sure if this really is correct.

phaethon:
Would it be a problem to set the limit for encoder to 8 symbols? For me it would make sense (before looking into code)
One thing that I realized - I have to implement prosigns and some international characters first on cw_gen side, for not to loose the functionality in this area. Which I anyway plan to do at some moment.

db4ple:
Eriks, I simply don't know. There is an argument in a comment in the CW_Decoder code which tries to explain why 40 has been chosen. I don't understand the comment to be honest. Maybe you can have a look as well. And yes, the code supports prosigns and some different languages codes. Most of this is disabled right now. So maybe we don't care too much at this moment about this.

PR #1096, opened 2017-10-01
Lcd experimental

sp9bsl:
Just for test that everything works.


db4ple:
Hi,
this pull request is not okay: it should have been against df8oe:lcd-experimental and not df8oe:active-devel.
@sp9bsl: You need to manually select the correct branch for the pull request. GitHub by default uses active-devel unfortunately and not the more logical lcd-experimental.

@df8oe: Why do you merge all commits from active-devel into lcd-experiemental? If Slawek does the rebase against df8oe:active-devel he will automatically integrate all changes from the main branch into lcd-experimental cleanly. No need to do this manually which clutters the history of lcd-experimental.
Let us try to use this approach.

df8oe:
@db4ple: Exactly that happened what I thought... sp9bsl has done his rebases to active-devel instead of lcd-experimental and so the pull request now is targeted to active-devel... My thought was if I synchronize active-devel and lcd-experimental he only has to wotk with lcd-experimental, rebases to this branch and nothing can go wrong... I will have a look if I can fix the wrong target via CLI git command...

df8oe:
@sp9bsl: It *is* possible to solve this manually on console. But I think it is better choice that you get more experience in using GitHub so I will not do it. I'll wait for the pull request that targets lcd-experimental...

@db4ple: we must discuss what branch is used for rebasing. If active-devel is used I must stop synchronizing branches here, if lcd-experimental is used I must synchronize in future, too. Problem is that both branches are present at Slaweks local repo and if he rebases from active-devel normally these targets his active-devel and not his lcd-experimental...

db4ple:
Hi Andreas,
Slawek should base his rebases on active-devel, that is what I told him. but he should target this pull requests to lcd-experimental. The main problem is GitHub, not Slawek. GitHub automatically assumes that you want to send a pull-request to the "main" branch. One has to actively tell GitHub to create the pull request against lcd-experimental. See the screenshot below: GitHub offers you to choose the target. Here the right selection has to be made.



You cannot fix the target with the git command line. It is not a git function, it is GitHub. Wait for Slawek to correct this problem. For him it is easy, just a new pull request.

db4ple:
And yes, the idea is to help Slawek to learn how to use GitHub. We all still have to learn how to operate the available infrastructure in the best way. This is part of the experimental approach to things. 
Lets wait for Slawek to fix the pull request and get on.
@sp9bsl: As you can see, we are trying to learn how to use GitHub and git with a concept of what is called a "shared feature branch". We have not done this in this way before, so we will have to see how it goes. I hope you don't mind this.

df8oe:
@db4ple: I _can fix it via git cli command_. But in this case it is not handled as a pull request :) I would have to pull the code from his repository, and manually add the changes to my lcd-experimental... If you take a look at the link "...or view command line instructions" right at each "Merge pull request" box you will get an idea how to do... I will not do it: but nice to know that it can be worked out... The power and flexibility of CLI git is amazing...

sp9bsl:
@db4ple i see that i put the cat among the pigeons :) All this is new to me and i'm still learning. I know where i made a mistake. I should push it to lcd_experimental, but i did it to master. I try again when i will be back at home after 17.00.
@df8oe i think i must do by the rules so if it is possible, remove please this pull request and i'll start again. Sorry for confusion.

db4ple:
@sp9bsl: As long as we all have enough fun, no problem. 
As you can see, a higher resolution display attracts some, so we should do it right. Of course, since we attempt to support both screen resolutions to keep existing users with the displays happy, this is a bit more difficult than just changing a few lines of code.  We should appreciate your efforts to do that extra mile and will have to wait (and support you as good as we can).

df8oe:
As I see pin layout of the LCD there are both data lines present. I cannot see if they are just connected to 2046 or to LCD controller.... There are different solutions for implementing both LCDs - but we should discuss this at issues not here at pull request...

No need to excuse. I have learned much during the time I supported this project and made many errors of course. They all have lead to the knowledge "way of community working with GitHub" I do have now. With patience and tolerance we all will find together :)

PR #1100, opened 2017-10-03
bugfixes and switch for LED

DD4WH:
- red LED, which is essential for right adjustment of the CW decoder threshold, can be switched OFF (not recommended)
- small bugfixes


DG9BFC:
switching red led off is senseless for me

DD4WH:
There were several complaints, that the LED is too bright and there should be a way to switch it OFF. Here it is :-).
If you have -once- adjusted your CW decode threshold to the signal and decoding works well, you can switch off the LED, if you want. Makes sense in some situations, I could imagine, does not make sense in other situations.

DG9BFC:
problem is that depending on signal you have to step in menue and change threshold (sometimes needs higher or lower setting) ... so for me the red led will stay on (yes you are right others may think different ... they could just desolder one leg of the led ... grin) ... if tooo bright ... install a series resistor in one of the led pcb traces (or wear sun-glasses ... hi hi)

db4ple:
I second @DG9BFC opinion. To use the CW decoder or CW Tune it is relevant to see the LED following the signal. If you don't like the brightness -> change the resistor, if you don't like the blinking at all -> you can't use the decoder or the tune helper, sorry.

PR #1103, opened 2017-10-03
lcd driver changed to use with ILI9486

sp9bsl:


mchf-eclipse/drivers/ui/lcd/ui_lcd_hy28.c
```diff
@@ -54,18 +54,23 @@
...
+#include "ui_lcd_hy28.h"
 
 mchf_display_t mchf_display;
 
 
+#ifdef USE_GFX_ILI9486
+//SP9BSL adapter board has CS pin on HY28B CSB pin
+#define LCD_CS					GPIO_PIN_12
+#define LCD_CS_PIO        		GPIOE
+#else
```
db4ple, line 67: We should place all concrete pin definitions in uhsdr_board_config.h not here, see below. Otherwise code cannot be easily ported to different board layouts if we have to hunt through all files to find out which pins are used where.
And in fact you are using the "normal" LCD_CS pin so there is already a definition available.
CSA is the CS pin for HY28A (which has a different pinout). Since we start testing for a HY28A SPI we initialize the cs pin first the with CSA pin data.

mchf-eclipse/drivers/ui/lcd/ui_lcd_hy28.c
```diff
@@ -605,29 +637,47 @@ uint8_t UiLcdHy28_SpiReadByteFast()
...
 void UiLcdHy28_WriteDataSpi( unsigned short data)
 {
     UiLcdHy28_SpiLcdCsEnable();
+#ifdef USE_GFX_ILI9486
+
+    GPIO_SetBits(LCD_RS_PIO, LCD_RS);
+    UiLcdHy28_SpiSendByte((data >>   8));                    /* Write D8..D15                */
+    UiLcdHy28_SpiSendByte((data & 0xFF));                    /* Write D0..D7   */
 
+#else
```
db4ple, line 664: Move the #ifdef so that it covers the real difference (the RS bit vs. the extra SPI byte).
In fact, we could simply call UiLcdHy28_WriteDataSpiStart. Much better.

mchf-eclipse/.settings/org.eclipse.cdt.managedbuilder.core.prefs
```diff
@@ -0,0 +1,13 @@
+eclipse.preferences.version=1
```
db4ple, line 1: I think we decided not to share this file. Can't remember the reason, to be honest. Might be that every Eclipse install will change this file and that causes lots of changes for nothing.

mchf-eclipse/drivers/ui/lcd/ui_lcd_hy28.c
sp9bsl: ok, i will change it.

mchf-eclipse/drivers/ui/lcd/ui_lcd_hy28.c
sp9bsl: Sorry, i don't understand

mchf-eclipse/.settings/org.eclipse.cdt.managedbuilder.core.prefs
sp9bsl: I suppose that this is by doing "git add . "

db4ple: git add . is an absolute no-go! I use git-gui (installed as part of Windows git) to control what I commit (I don't use the Eclipse integrated git support). This is much more convenient than the command line for that purpose. Most other stuff I use the command line but for preparing and managing the content of a commit git-gui is much better than the command line.

mchf-eclipse/drivers/ui/lcd/ui_lcd_hy28.c
db4ple: Final code:
 UiLcdHy28_SpiLcdCsEnable();
UiLcdHy28_WriteDataSpiStart();
UiLcdHy28_SpiSendByte((data >>   8));                    /* Write D8..D15                */
UiLcdHy28_SpiSendByte((data & 0xFF));                    /* Write D0..D7   */
UiLcdHy28_LcdSpiFinishTransfer();

No need for ifdef at all.

sp9bsl: OK i agree, it could be even shorter - without UiLcdHy28_SpiLcdCsEnable();  it is already in UiLcdHy28_WriteDataSpiStart();

PR #1105, opened 2017-10-03
Lcd experimental

sp9bsl:
Updated driver to support ILI9486


db4ple:
Hi Andreas,
this is the result of a rebase, no other changes so you can accept that. So far no problem.

But I see a problem coming: your merges somewhat destroy the idea of the rebase because these are also changing the "clean" history. I have not yet figured out if this will be a problem or not. Have to think about it.

PR #1106, opened 2017-10-04
Extending RIT limits to +-1200 Hz.

s53dz:


df8oe:
I never will use this because of never someone is more than 1khz "off" and if I want to do split operation _I_ use_split _operation_ - but it does not harm anything...

s53dz:
Andreas, I agree with you: 

> it does not harm.

TNX Bojan

PR #1107, opened 2017-10-04
Implementing prosigns for iambic entry

phaethon:
This extends iambic decoding and cw keyer to include prosigns. This can be basis to further add international characters.
This is, also, part of planned rework of cw decoder to use existing code in cw_gen for cw character decoding. In turn it will result in less used memory.

@db4ple I would be happy if you removed and/or explained more your plans regarding the hack in CwGen_AddChar. Adding prosigns made it a little more complicating. The prosign is enclosed in < and > and you should process it if it is not what you would want in rtty code. I am now using part of your hackcode myself for storing keyer macro memory. Please, see if just relying on ts.cw_text_entry would be enough for you. Intention of the variable is to indicate that iambic input should be processed as inputed text (like if from keyboard). One current issue is that there is no side tone if radio is not in CW mode.

### DIFF
--- a/mchf-eclipse/drivers/audio/cw/cw_gen.c
+++ b/mchf-eclipse/drivers/audio/cw/cw_gen.c
@@ -68,6 +68,8 @@
 // 3 => 8ms, 13 steps
 #define CW_SMOOTH_STEPS		9	// 1 step = 0.6ms; 13 for 8ms, 9 for 5.4 ms, for internal keyer
 
+#define CW_SPACE_CHAR		1
+
 typedef struct PaddleState
 {
 	// State machine and port states
@@ -106,7 +108,7 @@ static void   CwGen_TestFirstPaddle();
 // editors by placing both in vertically split window
 const uint32_t cw_char_codes[] =
 {
-		1,    // 		-> ' '
+		CW_SPACE_CHAR,    // 		-> ' '
 		2,    // . 		-> 'E'
 		3,    // - 		-> 'T'
 		10,   // .. 	-> 'I'
@@ -205,6 +207,33 @@ const char cw_char_chars[CW_CHAR_CODES] =
 		'-',
 		',',
 		':'
+};
+
+const uint32_t cw_sign_codes[] =
+{
+		187, //   <AA>
+		750, //   <AR>
+		746, //   <AS>
+		955, //   <CT>
+		43690, // <HH>
+		958, //   <KN>
+		3775, //  <NJ>
+		2747, //  <SK>
+		686 //    <SN>
+};
+
+#define CW_SIGN_CODES (sizeof(cw_sign_codes)/sizeof(*cw_sign_codes))
+const char* cw_sign_chars[CW_SIGN_CODES] =
+{
+		"AA",
+		"AR",
+		"AS",
+		"CT",
+		"HH",
+		"KN",
+		"NJ",
+		"SK",
+		"SN"
 };
 
 // Blackman-Harris function to keep CW signal bandwidth narrow
@@ -520,6 +549,7 @@ uint32_t CwGen_ReverseCode(uint32_t code)
 static void CwGen_CheckKeyerState(void)
 {
 	uint8_t c;
+	char prosign[2];
 
 	if (CwGen_DahRequested())
 	{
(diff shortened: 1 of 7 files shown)


phaethon:
The second commit removes a lot of code from cw_decode to use functions in cw_gen to decode symbols. We loose SOS prosign and international characters for now (which will come back at a later stage, when there is a good idea how to do them best).

df8oe:
Hi phaethon,
please take a look at this warning:
###################
../src/uhsdr_main.c:353:13: warning: passing argument 1 of 'strcpy' discards 'volatile' qualifier from pointer target type [-Wdiscarded-qualifiers]
      strcpy(ts.keyer_mode.cap[idx], "BTN");
             ^~
In file included from /home/travis/gcc-arm-none-eabi-6-2017-q1-update/arm-none-eabi/include/stdio.h:29:0,
                 from ../basesw/ovi40/Drivers/STM32F7xx_HAL_Driver/Inc/stm32f7xx_hal_def.h:50,
                 from ../basesw/ovi40/Drivers/STM32F7xx_HAL_Driver/Inc/stm32f7xx_hal_rcc.h:47,
                 from ../basesw/ovi40/Inc/stm32f7xx_hal_conf.h:244,
                 from ../basesw/ovi40/Drivers/STM32F7xx_HAL_Driver/Inc/stm32f7xx_hal.h:48,
                 from ../hardware/uhsdr_mcu.h:26,
                 from ../hardware/uhsdr_board.h:16,
                 from ../src/uhsdr_main.c:16:
/home/travis/gcc-arm-none-eabi-6-2017-q1-update/arm-none-eabi/include/string.h:34:8: note: expected 'char * restrict' but argument is of type 'volatile uint8_t * {aka volatile unsigned char *}'
 char  *_EXFUN(strcpy,(char *__restrict, const char *__restrict));
###################
You can see those warnings in the build log of Travis which is available some minutes after you have created pull request. If your systems warnings level is different from mine maybe you do not see all warnings.

df8oe:
I merge now - but please take care that there are no warnings at building ;)

mchf-eclipse/drivers/audio/cw/cw_gen.c
```diff
@@ -699,13 +815,42 @@ void CwGen_AddChar(ulong code)
...
+		for (int i = 0; i<CW_SIGN_CODES; i++)
+		{
+			if (cw_sign_codes[i] == code) {
+				result = '-';
+
+				UiDriver_TextMsgPutSign(cw_sign_chars[i]);
+
+				// FIXME: HACK HACK FOR RTTY TX TESTING
+				// We can use the same buffer for all digimodes
+				if ((ts.txrx_mode == TRX_MODE_TX && is_demod_rtty()) || ts.cw_text_entry) // Can rtty rely just on cw_text_entry here?
```
db4ple, line 839: The idea is, check if a mode is active which accepts CW based input during TX, is active. Right now this definition applies to RTTY only. To scale this, I would suggest to either use ts.cw_text_entry for that purpose, i.e. any mode which wants text entry has to set this when it is going to  tx. Or we write a function to tell us this dynamically. Or we use a single additional variable to distinguish between a text entry mode for things like recording buttons or menu items and the use during transmissions as direct source of input.

mchf-eclipse/drivers/audio/rtty.c
```diff
@@ -717,6 +717,14 @@ int DigiModes_TxBufferPutChar(uint8_t c)
     return ret;
 }
 
+void DigiModes_TxBufferPutSign(const char* s)
```
db4ple, line 720: Why not process as strings here, your input in your datastructure are strings anyway.

mchf-eclipse/drivers/audio/cw/cw_gen.c
phaethon: I would see ts.cw_text_entry as a single variable to check in this if. All modes, which allow such input then set its value to true and are responsible for setting to false after exiting the mode. Makes interface cleaner.

phaethon:
It could be processed as string, but current implementation allows the argument being 2 character array, which does not end with \0. Might make some things a little easier and less error prone for future use cases (for the cost of limiting to 2 character prosigns).

Issue #1049, opened 2017-09-09
E2 second function - forced in RTTY when cycling modes


s53dz:
When cycling thru modes (using G1 (S16)) and when passing the RTTY mode the M2 (S3) switch switches to its second function (RTTY SHIFT setting) and stays there.

(FW D2.5.82, but probably also before that.)

73 Bojan

db4ple:
Yes. This is a  "work in progress" area since the old code lacks a little in that department (flexible switching encoder layouts between modes was all handcrafted code. This gets with each mode more difficult to maintain and change).

s53dz:
Danilo,

I understand it is not easy to maintain all settings between switching modes, that is why I started this issue in an easy way: if it is possible just to remove pushing the second function when in RTTY mode it would release this issue then.

What happens is this: You switch modes across and end-up in SSB and try to change AGC / RF gain, but you are changing the RTTY shift. To get back to AGC you have to roll thru a few of E2 functions.

(D2.5.83)

EDIT:
Additional information: When changing between digital modes on a TouchScreen E2 function stays (which is ok).
(D2.5.87)

EDIT2:
When digital mode is FreeDV the E2 function does not change when cycling thru modes (which is ok). It happens only when chosen digital mode is RTTY.
(D2.5.92)

73 Bojan

PR #1117, opened 2017-10-09
Fix for the issue 1049.

s53dz:
RTTY E2 second function only forced when not in RF gain mode.


db4ple:
Hi Bojan, just a hint: placing a hashmark in front of the issue number makes it a clickable link: #1049
And automatically creates a backlink in the issue to the comment or pull request.

PR #1118, opened 2017-10-09
Changed waterfall display memory mapping

sp9bsl:


sp9bsl:
image

mchf-eclipse/drivers/ui/lcd/ui_lcd_hy28.h
```diff
@@ -24,6 +24,13 @@
 	#define MAX_Y  240
 #endif
 
+#ifdef TimeDebug
+#define MARKER LCD_D0
+#define MARKER_PIO LCD_D0_PIO
+
+#define Marker_ON  MARKER_PIO->BSRR=MARKER;
```
db4ple: Hi Slawek,
please do not use direct assignments, these are not portable (your code works not on F7 and F4 not on H7, ask STM why).
GPIO_SetBits() / GPIO_ResetBits() are our portable macros.

Using a "normal" GPIO is also not recommended since it is also not very portable, typically the green or red led should be used since available on all platforms.

sp9bsl: OK, i'll remove this code after we end with lcd  optimization.

sp9bsl: Danilo, what is wrong now? I see in conflicts that it refuses all changes i made.

db4ple:
Hi Slawek,
I cannot see the conflicts unfortunately. As before, please ignore the conflicts, these are (in this case) not your problem.

df8oe:
I have merged manually.

PR #1121, opened 2017-10-10
Changes to support STM32H7 Eclipse builds

db4ple:
No functional changes for other builds.

No support for makefile builds yet, no actual CubeMX H7 files checked in,
just the configuration.
H7 Linker script for start without bootloader


df8oe:
Very interesting!! May it be that I can test H7 with bootloader next days?? LCD output ;) ?? You are working quite fast...

PR #1126, opened 2017-10-13
Support files for mcHF and OVI40

K2AAE:
This is the support files for mcHF and OVI40: H7 F7 SD and UFS memory, Eeprom, Flash, Xtal , Codecs, Stereo and Power jacks.


db4ple:
Hi Kevin,

glad that you are looking to complete our datasheet collection!

Please follow a few (not yet explicit documented) rules:
- Use a uniform naming scheme for the documents (Datasheet_<HW_Item_Name>). I know, the last commit did not follow that scheme. @df8oe: Andreas, 👎 for that :-)

- Only include software related hardware items (the idea here is to collect the information actively used to develop the firmware, not a full hardware design documentation). Design for jacks do not fit here. Similarily I don't think we need the Nucleo Board documentation.
 
- Only include documents which are actually being used (do not commit documents for devices no one is working on (yet)). These should be included when someone works on this in their respective pull request. And you have all the rights to complain about missing documents at that point in time.

df8oe:
Hi Kevin,

thank you for your contribution! As Danilo stated I will not merge this at the actual state because there are many files which are related to parts that are not used at mcHF or OVI40 and/or there is no intention to implement them in the next future. A naming scheme as proposed by Danilo will help to quick identify where a document is related to and so is appreciated. I am looking forward to a pull request that will follow these direction and of course will merge it.

K2AAE:
Bye

PR #1127, opened 2017-10-13
Implementing BPSK modulation #1002

phaethon:
Demodulation is still in works, but somebody if interested can test out transmission meanwhile.


df8oe:
This does not build (see travis-ci output). Please take a look at that again.

phaethon:
On eclipse it did build. Added to files.mak and f7-files.mak, and now is ok with Travis, too. 
btw, it is not nice that the file needs to be added manually to both files. Would be better to have files.mak for common files for all platforms, and then f4-files.mak and f7-files.mak for platform specific.

db4ple:
Yes, you are right re two files. But on the other hand, it is not too much work. Wait until we have the H7, then you can add that to three files ;-)

I think we should fix that. This would even make the naming scheme a little more logical...

phaethon:
I hardly imagine how we will assure reasonable testing across multiple hardware platforms. I naturally test only on what I have (which is original mchf with f4).

db4ple:
Well, that is naturally going to be a challenge.

db4ple:
Hi Eriks,
I would factor the F4 specific files out (into f4-files.mak) and leave the common files in files.mak  in one of my next pull requests. Or do you want to do this for the makefile builds?

phaethon:
Please, do it.

PR #1136, opened 2017-10-19
Stereo demodulation modes prototype

DD4WH:
Implemented stereo demodulation for the OVI40 hardware (does NOT work on mcHF, sorry)

- switch on STEREO demod in Debug menu
- choose between 
* IQ demodulation (I right ear, Q left ear) or
* STEREO SSB (LSB left ear, USB right ear)

several ToDos:
* implement STEREO SAM demodulation --> DONE
* lattice filter crashes when switching to CW (therefore disabled at the moment) --> FIXED
* automatic LMS notch filter and LMS noise reduction cannot be used in stereo mode (yet)
* graphical indicator for filter bandwidth has to be adjusted to cover both sidebands --> FIXED

Highly experimental, therefore no WIKI entry or forum announcement yet, use at your own amusement


mchf-eclipse/drivers/audio/audio_driver.h
```diff
@@ -74,6 +74,8 @@ typedef struct
     float32_t                   a_buffer[IQ_BUFSZ];
     float32_t                   b_buffer[IQ_BUFSZ];
 
+    float32_t					r_buffer[IQ_BUFSZ]; // used for the right channel in STEREO DEMODULATION
+
```
db4ple: If not used in mono config, please ifdef

mchf-eclipse/drivers/audio/audio_driver.c
```diff
@@ -1993,10 +2132,36 @@ void AudioDriver_RxAgcWdsp(int16_t blockSize)
             agc_wdsp.in_index -= agc_wdsp.ring_buffsize;
         }
 
-        agc_wdsp.out_sample[0] = agc_wdsp.ring[agc_wdsp.out_index];
+//        agc_wdsp.out_sample[0] = agc_wdsp.ring[agc_wdsp.out_index];
+        agc_wdsp.out_sample[0] = agc_wdsp.ring[2 * agc_wdsp.out_index];
```
db4ple, line 2140: You sure that this applies 2x to non-stereo config as well? Just asking.

mchf-eclipse/hardware/uhsdr_board.h
```diff
@@ -1021,6 +1024,7 @@ typedef struct TransceiverState
 
 	keyer_mode_t keyer_mode; // disable/enable keyer mode for F1-F5 buttons
 	bool buffered_tx; // disable/enable buffered sending for CW and digital modes
+	bool stereo_enable; // enable/disable stereo demodulation (only in special hardware, NOT in mcHF)
```
db4ple: Why not ifdef it not usable on all hardware?
Allows us to clearly identify all related places for hardware specific features. And the compiler will tell you where you make references to stereo_enable in "mcHF" code, where it is not supported at all.

mchf-eclipse/drivers/ui/radio_management.c
```diff
@@ -1133,6 +1133,18 @@ bool RadioManagement_IsApplicableDemodMode(uint32_t demod_mode)
...
+    case DEMOD_IQ:
+    	if(!ts.stereo_enable)
+    	{
+    		retval = false;
+    	}
+    	else
+    	{
+    		retval = true;
+    	}
+    	break;
```
db4ple, line 1147: retval = ts.stereo_enable would do the same job in a single line ;-)

db4ple (changes requested):

mchf-eclipse/drivers/audio/audio_driver.h
DD4WH: I now defined different AGC function calls, one with one pointer, the other with two pointers = stereo, depending on the hardware --> r_buffer is only used in STEREO modes

mchf-eclipse/drivers/ui/radio_management.c
DD4WH: Oh, thanks! Acknowledged ;-). Should I change it or will the compiler do the job of simplifying?

mchf-eclipse/hardware/uhsdr_board.h
DD4WH: Done.

DD4WH:
Would be good if somebody checks whether this still works on an mcHF.
I did check on UI40 board with RF board mcHF 0.4. That works nicely.

df8oe:
Hi Frank,

mcHF seems to work (I cannot test every function and combination). But the new "modes" SSB-S" and "IQ-S" appears on mcHF. Is that correct?

73
Andreas

DD4WH:
thanks, Andreas, for testing!
No, they should not appear in the mcHF. I think I already found the bug and will commit a fix this evening.
Would be good if you wait with the merge until the fix.

just for me a reminder:
* uhsdr_board.h should have the DEMOD_MAX_MODE dependent on #ifdef USE_TWO_CHANNEL_AUDIO
* ui_driver.c line 1522 insert #ifdef
* ui_menu_internal.h line 239: put #ifdef

df8oe:
now I do nit see any issue - let's look if others see some ;)

PR #1153, opened 2017-10-27
leaky LMS noise reduction & bugfix SAM

DD4WH:
- bugfix in fade leveler for AM/SAM RX
- cleanup of no longer used parts of the snap algorithm
- implemented leaky LMS noise reduction: adapted from WDSP lib
Leaky LMS noise reduction could be more effective and less annoying than the (already implemented) standard LMS noise reduction. You have to try it out. I find it slightly more comfortable for the ear.
LMS noise reduction is only useful in cases where there is already a good SNR. It cannot draw tiny little signals from the noise.
You can compare the standard LMS noise reduction and the leaky LMS noise reduction by a switch in the debug menu ("leaky LMS"). There are also some options to finetune the algorithm, although it is not so obvious in the audio (at least during my tests). The leaky LMS needs almost 70% more processor power than the standard LMS, so this could be limiting with the STM32F4. At the moment, in leaky LMS, you cannot produce nice audio when you have simultaneously switched on noise reduction NR AND automatic notch filter. But leaky LMS noise reduction alone and leaky LMS notch filter alone both work well.
I am not yet really convinced that we need both noise reductions, however there have been users (Sigi ! :-)) who wanted to have more possibilities to finetune their noise reduction according to their needs. This is possible with the leaky LMS. 
I have not tested whether the leaky LMS runs on STM32F4 machines.


DG9BFC:
produces nice audio .. runs fine on f4 ... but i have to crank up audio level to high niumbers to have same level as with old nr ... so not easy to compare (when set to comfortable audio and switched off the speaker jumps of the table!)... the settings are doing almost nothing (especially the gain settings in leaky lms!) ...

PR #1156, opened 2017-10-28
Si5351a improvements

db4ple:
Hopefully correct handling of limits
automatic switching from quadrature to single clock based on requested frequency.
This is to be changed in the final system but helps testing.
Seems to be sufficiently stable also in the 2m band.

### DIFF
--- a/mchf-eclipse/drivers/ui/oscillator/ui_si5351a.c
+++ b/mchf-eclipse/drivers/ui/oscillator/ui_si5351a.c
@@ -20,12 +20,13 @@
 #include "uhsdr_hw_i2c.h"
 #include "ui_si5351a.h"
 
+#ifdef USE_OSC_SI5351A
 // Will be removed and made a dynamic config element
 #define TEST_QUADRATURE
 
 // reference oscillator xtal frequency
 #define SI5351_XTAL_FREQ		27000000			// Crystal frequency
-#define SI5351_MIN_PLL			600000000			// Min PLL frequency
+#define SI5351_MIN_PLL			405000000			// Min PLL frequency
 #define SI5351_MAX_PLL		    900000000			// Max PLL frequency
 
 // Register definitions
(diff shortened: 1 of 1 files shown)


S52CQ:
Hi,
I am reading msgs. regarding Si5351a. Does that mean that you guys are implementing Si5351a into mcHF?
Would that mean more coverage of mcHF? Is it possible to test that? I've got few Si5351a in my drawers so I could easily test it.
73s de Jure

df8oe:
Hi Bojan,

SI5351A will be the LO of OVI40 so development is done for that rig. Of course SI5351A will work in mcHF, too because it is autodetected. BUT: Hardware of mcHF is complete different from OVI40 so it will not work as expected without firmware or hardware modifications of mcHF.

73
Andreas

S52CQ:
Hi Andreas,
I'd love to make that modification to replace SI570. Do I just connect SCL/SDA and Clock output instead of SI570? Which Output on SI5351A did you use to be active? I presume that I have to add crystal to
SI5351A.  Will there be more freq. coverage than SI570 - right?

73 de Jure, S52CQ

db4ple:
Hi,
a) the Si5351a code is not ready for "production" use. It is experimental.
b) it is possible to replace the Si570 and yes, there would be somewhat higher frequencies possible. 

But of course it requires some more effort -> BPF, LPF needs to be changed, software has to recognize that additional bands can be used etc.  The while the UHSDR has support for 6 and 4m bands  it would not know that your mcHF supports these. So it is a longer path if the Si5351a is going into main stream. For just playing with it, the changes are rather trivial but these then cannot go into the main code base with much extra work if we want this "dynamically" switchable. There is the opportunity to define a new "hardware" and make this a compile option. That would be possible which much less work.

73
Danilo

S52CQ:
Hi Danilo,

tnx for reply. Actually I intend to use SI5351a in mcHF for covering freq. below 1.8 MHz, down to a couple of hundred kHz. I guess this should be done without any major mods to mcHF right?

73 de Jure

df8oe:
At OVI40 SI5351A is used in two different modes which require hardware switching. You get fx4 if you are below a "magic frequency" and if you are about that you will get f (but with 90° phase offset on CLK0 and CLK1). The hardware for this is not present at mcHF. So if you want to use it as a "straight oscillator" you must take a look at the code and set the "straight mode" always on.

But if you leave the band limits (and 160m is already "out-of-BPF") you will get very poor results. You have to add more BPFs. That will be enough for only RX. If you do want to TX you need LPFs, too.

S52CQ:
Or should I just wait for OVI40 to be available :)

BTW - when do you plan to make it available for the public?

73 de Jure

df8oe:
UI-PCB will be available end of the month (and can firstly be used with mcHF RF board). Hopefully OVI40 RF-PCB will be available at end of this year.

73
Andreas

PR #1193, opened 2017-11-20
Fix for broken (Eclipse/embitz) builds if no external TRX_ID / TRX_NAME defines are used.

db4ple:


df8oe:
I wa simultaneously working on this...

PR #1194, opened 2017-11-20
Refactored display init code / OVI40 display fix

db4ple:
- more streamlined init code, using a data structures to read configuration
- fixed regression with OVI40 white screen in last few commits
- Should work for OVI40/mcHF ILI932x and ILI9486 displays
Tested OVI40/ILI932x par, mcHF ILI932x SPI, mcHF ILI9486 SPI

I noticed a broken OVI40 parallel display after rebase. Need to check that.

PR #1195, opened 2017-11-20
SPI speed-up for mcHF with ILI9486

sp9bsl:
Wrong SPI init function call parameter fixup. After touchpanel transaction (for example detection of presence) there was not proper prescaler value set (division by 4 instead of 2). This will not affect the OVI40 SPI mode.

### DIFF
--- a/mchf-eclipse/drivers/ui/lcd/ui_lcd_hy28.c
+++ b/mchf-eclipse/drivers/ui/lcd/ui_lcd_hy28.c
@@ -1826,7 +1826,7 @@ uint8_t UiLcdHy28_Init()
     {
         UiLcdHy28_ParallelDeInit();
     	mchf_display.use_spi = true;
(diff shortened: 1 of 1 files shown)


db4ple:
Hi Slawek,
please review my changes in #1194  (did change a lot in the ui_lcd_hy28.c) and also fixed the problem with the prescaler. Maybe F7 SPI prescaler is not handled correctly, but I doubt your patch will work on the changed code.

sp9bsl:
Hi Danilo,
yes you're right. We have doubled :) With this array, the calling parameter for initSPI is ok. There is only needed the workaround for slowing down the spi for F7. If it is not set to divide by 4, Andreas will have white screen with parallel. I will check this with OVI board when i solder it.
Issue with the prescaler i found at spring, but forgot to add it to new code.

PR #1199, opened 2017-11-22
Reordering and small changes and fixes in LCD init

db4ple:
- Concentrated controller specific code (less #ifdefs)
- More generalized GPIO init code
- with fixed FMC_Init/FMC_DeInit we can now call Parallel_DeInit() in the correct place
- fix for spi speed selection.
No real functional changes.


db4ple:
Both if used with single controller driver and with multiple controller driver?
All other displays@OVI40/mcHF work or not tested?

PR #1201, opened 2017-11-23
Spectral noise reduction

DD4WH:
Alpha version of spectral noise reduction
Implementation of spectral weighting approach following Schmitt et al. 2002 using the Romanin et al. 2009 simplified noise estimator together with the voice activity detector of Hu et al. 2002.
For a detailed explanation of all this stuff have a look at the WIKI.
https://github.com/df8oe/UHSDR/wiki/Noise-reduction

Enable this Noise Reduction in the Debug menu.
Proposed parameter settings (in Debug menu):
NR alpha - 800 to 980 (the higher - the stronger the noise reduction effect)
NR thresh - depends on your noise level - 200 to 20000
NR beta - strongly recommended to leave this at 850
Please test and report !
Developed together with Michael - DL2FW - thanks !!!


ly2bok:
Something is not right. This function does not work! The sound is not clean, with interruption. It is necessary to check ....
LY2BOK

DD4WH:
Hi,
thanks for testing, yes, the function does not seem to be initialized in the right way. Will try to fix that in the next days.
However, I never had interruptions in the audio, that is strange!? Did you enable it with the single switch in the DEBUG menu? DO NOT enable the standard NR.
73 Frank

DD4WH:
BTW, which hardware do you use?

ly2bok:
NOT enable the standard NR.
SDRmcHF v.4

PR #1205, opened 2017-11-23
Spectral noise reduction repair

DD4WH:
* now a considerable noise reduction effect is achievable
* I can hear some slight distortion on speech signals --> has to be tweaked
* try NR alpha = 0.95 and thresh = 8000
This is still an early version, work in progress, you can try it out, but we know it has to be improved in sound quality.

### DIFF
--- a/mchf-eclipse/drivers/audio/freedv_uhsdr.c
+++ b/mchf-eclipse/drivers/audio/freedv_uhsdr.c
@@ -754,10 +754,10 @@ static float32_t NR_last_iFFT_result [NR_FFT_L / 2];
 static float32_t NR_last_sample_buffer_L [NR_FFT_L / 2];
 float32_t NR_FFT_buffer[NR_FFT_L * 2];
 float32_t NR_iFFT_buffer[NR_FFT_L * 2];
-static float32_t NR_X[NR_FFT_L / 2][2]; // magnitudes of the last two FFT results for 128 frequency bins
-static float32_t NR_M[NR_FFT_L / 2][2]; //
-static float32_t NR_Gts[NR_FFT_L / 2]; // time smoothed gain factors for each of the 64 bins
-static float32_t NR_G[NR_FFT_L / 2]; // preliminary gain factors (before time smoothing) and after that contains the frequency smoothed gain factors
+static float32_t NR_X[NR_FFT_L / 2][2]; // magnitudes of the current and the last FFT bins
+static float32_t NR_Nest[NR_FFT_L / 2][2]; // noise estimates for the current and the last FFT frame
+static float32_t NR_vk[NR_FFT_L / 2]; //
+static float32_t NR_Hk[NR_FFT_L / 2]; // gain factors
 static float32_t NR_SNR_prio[NR_FFT_L / 2];
 static float32_t NR_SNR_post[NR_FFT_L / 2];
 static float32_t NR_SNR_post_pos[NR_FFT_L / 2];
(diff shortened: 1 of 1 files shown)


ly2bok:
Nothing good .... listen to the examples.

http://www.qrz.lt/ly2bok/Laikinas/

df8oe:
I am listening to SSB since some hours and comparing again and again recent NR and it is working absolutely great. Cannot confirm sounds like on your recordings. My version is 2.7.11. Do you use "Spectral NR" from Debug menu? That's the one we are talking about...

ly2bok:
Yes , version  is 2.7.11.  I use "Spectral NR" from Debug menu. SDRmcHF version 4.Very strange...

df8oe:
There must be some other settings at your mcHF which do eat much horse power. I do not know...

I by myself are always running in magnify x 1, moderate waterfall (no scope), filter bandwidth 2.7K or 2.9K (in SSB). And my mcHF is also 0.4.

df8oe:
I think I have found the reason: all of my mcHFs do have STM32F429/439 MCUs which do have 256KB RAM instead of 192KB... Do you have 192KB MCU?

ly2bok:
Yes,  work is STM32F405.I understood.
I think that if you publish the program, then it is for everyone. And now it turns out only for those who transferred SDRmcHF to the new STM. We need to inform about this and do it in a new section (only for new STMs) Now we get confusion.
Thank you  , Good luck.

DD4WH:
Hi ly2bok, Thanks a lot for your tests!

Hey, come on, its an early test version :-).

Nothing is "published", this is TESTING. But you are right, its for everyone! If you would like to have thoroughly tested software, take the released versions. But with these dailys, there is no documentation apart from the info on github. 

Don´t get confused, it is quite easy. We are working on this topic constantly, and IF it is possible that this runs on the small machines, we will inform you. IF it is not possible, then you have to live with that.

Everyone is invited to test with us these daily versions, but we will not guarantee that it works on your specific machine. But we do everything we can to MAKE it work (and for that we need YOUR feedback from the STM32F405, so please continue to report if it´s not working  for you!).

So, let´s work and improve! :-)

73 Frank DD4WH

DD4WH:
BTW, it could be that it now works on your machine with the very latest daily, because Danilo and Michael and me improved the RAM usage . . . would be nice to hear about your latest test reports with the newest firmware version ;-)

df8oe:
For your information:
UHSDR is *not* a firmware only for mcHF. Of course we want to support mcHF as good as possible. But due to hardware lacks on mcHF some of the features will not work on mcHF because hardware which is used there is not capable to support all new features. Actually these features do not run on mcHF:
- stereo output (mcHF only does have one audio PA)
- VOX (we are working on it) - (mcHF only does have one audio codec)
and there will be much more in future! Some can be implemented by swapping small MCU to STM32F429/439 - if RAM is the bottleneck. If you do have a 405/407 with only 512KB Flash this is a bottleneck, too and can be removed by swapping MCU. The facts that firmware development is at very high speed and will need a minimum 1MB flash MCU and better 256KB RAM is known since nearly 2 years. But we are not the maintainer of the hardware - that is Chris M0NKA. We are only working on UHSDR firmware which will support more than mcHF (see GitHub readme.md). Original mcHF firmware which is only capable to run on mcHF is available at the web pages of Chris. But it seems the development has stopped.

73
Andreas

ly2bok:
.I understood. Regrettably.
Thank you.

PR #1208, opened 2017-11-23
Spectrum FFT size has been made runtime switchable

db4ple:
Default configuration now uses 320x240 but includes the larger
fft capabilities to see if this causes issues.
Please note: This is an internal capability for now, users cannot switch the FFT size.
We are going to use this to support different display resolutions with a single firmware binary.


df8oe:
First review: mcHF @320x240 works well, OVI40 @480x320 does only show black areas where wf and scope should be. Build using Makefile with option LCD_TYPE=1

PR #1210, opened 2017-11-24
New in-place FFT width to spectrum width algo

db4ple:
No more need for extra buffer, please review


sp9bsl:
Hi Danilo,
for a first glance:
```
if(from_len==to_len)
 return;
```

More at evening, no time now sorry :)
73 Slawek

Issue #1191, opened 2017-11-20
Support for SPI DMA @OVI40 UI LCD


df8oe:
We will need SPI DMA on other MCUs as F4 in the future, too...

df8oe:
implemented.

PR #1217, opened 2017-11-25
 Added Display SPI DMA support for OVI40 / F7

db4ple:
Please note: This needs to be tested. Configurations not using
OVI40 UI with SPI display should not be affected.

That is the intended solution for #1191


df8oe:
I can confirm that SPI DMA now does not lead to LCD hanging. It is not as fast as "it should be" - but it is faster than before of course.

db4ple:
Hi Andreas,
you can try if your display runs with SPI Prescaler 2 (instead of the "safe" high speed prescaler 4). See ui_lcd_hy28.c around line 579.

db4ple:
And don't expect wonders from the SPI DMA. The perceived slowness of the waterfall in 480x320 btw is just that: perceived. With some changes, the waterfall is quite fast even on a F4 (using SPI DMA of course).

df8oe:
Hi Danilo,

of course, I do not expect wonders :) My display does not work with prescaler set to 2 - only white screen. And there is no need to improve SPI LCD on OVI40, because all LCDs will be driven parallel (I hope). But having SPI DMA is a good-to-know...

sp9bsl:
I must add one thing, when I switched to parallel I don't want to go back to serial, even with the prescaler_2 :) But i know that there will be someone that wants to do it...

df8oe:
I don't think anyone will use LCD in SPI on OVI40. But regarding mcHF there will be some because of a lack of free GPIOs. So improvement of SPI speed on mcHF is an important point.... Not very high priority, but it is there.

sp9bsl:
Agree, I will not give it up. Must try my old idea with predrawn buffers, for spectrum draw. There is a lot of calculations and checking for draw few pixels per bin. With waterfall there is nothing more we can do than we have today, simply the SPI is a bottleneck.

db4ple:
Hi Slawek,
before you start optimizing, you may want measure the actual computation (and drawing time) for parallel and spi.
If both are similar -> computation is the problem, if spi is much slower, the actual transmission is the bottleneck. And if the total time for all of that is low, why bother to optimize something not taking a lot of time at all...
Just my thoughts. There is profiling.h which gives you some "tools" to measure.

sp9bsl:
Yes, I know and always do that before try to optimize something, I will try it but not now, I think the unification have higher priority. As I remember there was huge difference in 2.2.0 when i did it. Today it is not clear because the code is different. Well, we will see after some tests :)
And more, if there is nothing that is worth of optimizing, the only way to speed it up is to change the waterfall line count (maybe another menu intem for change it). Some people asked me for this.

sp9bsl:
Few days ago Andreas gave me the idea: there should be waterfall that it's speed changes progressively (slows down in few steps) - as i remember it is used in quisk. Maybe this is the idea which we someday implement? It should decrease the load due to rare refresh of old data.

Issue #1197, opened 2017-11-21
SCOPE NoSignal ADJ


WM0I:
Not working. Set values from 10 to 30 with no effect on display. Am I missing something?

df8oe:
This unused setting is gone now.

db4ple:
Ups, I did not notice that we had an issue for that...
So now to link this to #1221

PR #1221, opened 2017-11-26
Refactored Scope/Waterfall to be prepared for multi resolution binaries

db4ple:
- 480x320 resolution -> Waterfall and Scope have now independent speed settings
- 320x240 resolution -> minor layout changes and fixes
- disabled unused scope adjust setting, which fact was issue #1197
- reorganized location of spectrum / scope / waterfall settings
- on STM32F4 480x320 waterfall always uses line doubling to save memory.
- waterfall speed can be now set to "OFF" to stop waterfall


sp9bsl:
Hi Danilo,
why do you define screen coordinates outside the ui_lcd_items.h?
lines 44 and 45 of ui_spectrum.c

73 
Slawek

PR #1232, opened 2017-11-29
spectral noise reduction refactoring: variables in struct

DD4WH:
* variables for spectral noise reduction now defined in audio_nr.h as a proper struct
* added code for debug display of bin gain values and gain values for long tone detection 
* DEBUG menu: Show gains:

Hk: this the gain that is applied for the specific frequency bins for noise reduction
notch: this is the gain applied to bins where long tones have been detected
Hk & notch: the two gains are multiplied and the result of the multiplication displayed

BEWARE: the display shows GAINS, not any kind of SIGNAL. This only shows the multiplication factor, that the signal is treated with during the NOISE REDUCTION and AUTO NOTCH (=long tone elimination) process


df8oe:
does not build due to errors - please look at Travis...

DD4WH:
very instructive error message indeed: 
"The command "make -f ../Makefile ROOTLOC=".." LCD_TYPE=1 all" exited with 2."
Cannot interpret this at all, please talk English to me ;-).

db4ple:
Hi Frank,
you have to scroll up, directly above the cited message are the compiler error messages.
And it tells you exactly what the problem is: DO NOT ADD __MCHF_SPECIALMEM in header files, only in the  one C file where the variable is defined.

df8oe:
...it's not because of Travis - this is an unallowed way of programming. We have screwed down warning level and have made Travis very informative to find "hidden issues" whose warnings are eaten by high warning level... So Travis is our very good friend :)

mchf-eclipse/drivers/audio/audio_nr.h
```diff
@@ -24,18 +24,18 @@
 
 typedef struct NoiseReduction
```
db4ple: Hi Frank, this is not a good idea. Generally a struct is a good thing. In our case, we need two for the sake of memory allocation. 
A struct is always placed consecutively in memory, that means either ALL or NOTHING is in the special mem.  Due to the tight memory situation, we need to push as much as possible of this data in the special mem. But not all of it does fit into there. That is why not all had the __MCHF_SPECIALMEM in front of them when they were single variables.
We need to roll this back and revise your idea, otherwise we will have the memory issue again on the small machines.

DD4WH:
Hi Andreas, hi Danilo,
thanks for the explanation!
I will not start to love travis, "has exited with 0" means "everything is OK" --> who has invented this highly misleading and irritating nonsense? That is ridiculous ;-)

DD4WH:
OK. Now I understood. What I will do now is have two structs and one is in mchf special memory, the other isnt,
I will allocate exactly those variables into the special / non-special memory where they were.

db4ple:
This is Unix -> return 0 means (as a convention) all is good. Any other number is often considered as an error and you would look up the error code in the manual (in the old days, e.g. the 60th/70th).

DD4WH:
Thanks! 
Will push better code in the next hour :-).

db4ple:
Yes, exactly 2 structs. From a "software purist" point of view this is bad, but is a pragmatic solution, since it is way better than no struct. Anyway, good that we talked about it :-)

df8oe:
Yes - "0" means "no error".

But Travis does many steps and all of them returned "0" until there is one that errored - and that returns not "0". It is standard way in Linux scripting to chain many different steps to a script and if you let it log the results, you will see that if averything is ok each stept returns "0" and if one step does not return "0" script stops at this line... So you have to pay attention to look at the errored part - therefore you can rollback the output of Travis and see all errors (and also warnings!!) that ocur during Makefile building. Simple, pragmatic and efficient ;)

PR #1233, opened 2017-11-29
spectral noise reduction - refactoring

DD4WH:
* all NR variables are now arranged in two structs
* Variable allocation to memory (mcHF_special_memory and the other) should now be exactly the same as before

### DIFF
--- a/mchf-eclipse/drivers/audio/audio_nr.h
+++ b/mchf-eclipse/drivers/audio/audio_nr.h
@@ -22,7 +22,7 @@
 
 #define NR_FFT_L NR_FFT_SIZE
 
-typedef struct NoiseReduction
+typedef struct __MCHF_SPECIALMEM NoiseReduction
 {
 	float32_t 					last_iFFT_result [NR_FFT_L / 2];
 	float32_t 					last_sample_buffer_L [NR_FFT_L / 2];
--- a/mchf-eclipse/drivers/audio/audio_nr.h
+++ b/mchf-eclipse/drivers/audio/audio_nr.h
@@ -28,7 +28,6 @@ typedef struct __MCHF_SPECIALMEM NoiseReduction
 	float32_t 					last_sample_buffer_L [NR_FFT_L / 2];
 	float32_t 					Hk[NR_FFT_L / 2]; // gain factors
 	float32_t 					FFT_buffer[NR_FFT_L * 2];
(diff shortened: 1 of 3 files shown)


df8oe:
If you take a look at Travis for this commit you can see that there are some warnings... I do not know if you can take a look at older Travis logs, too or if this is only possible for repository owner...

DD4WH:
I am not sure whether the warnings mean that the assignment of the struct to mchf_special_memory is not valid? 
"In file included from ../drivers/ui/ui_driver.c:29:0:

../drivers/audio/audio_nr.h:40:1: warning: 'section' attribute does not apply to types [-Wattributes]

 } NoiseReduction;"

Danilo, can you tell us more about this?

db4ple:
Sure, I can, once I looked at it.

mchf-eclipse/drivers/audio/audio_nr.h
```diff
@@ -28,26 +28,35 @@ typedef struct __MCHF_SPECIALMEM NoiseReduction
 	float32_t 					last_sample_buffer_L [NR_FFT_L / 2];
```
db4ple, line 28: __MCHF_SPECIALMEM does not belong to a type definition (as the compiler told you).
A type in C does not have a memory location.
See below for next comment.

mchf-eclipse/drivers/audio/audio_nr.h
```diff
@@ -28,26 +28,35 @@ typedef struct __MCHF_SPECIALMEM NoiseReduction
...
 	float32_t 					SNR_post[NR_FFT_L / 2];
 	float32_t 					SNR_post_pos; // saved 0.24kbytes
 	float32_t 					Hk_old[NR_FFT_L / 2];
 	float32_t 					VAD;
+	int16_t						gain_display; // 0 = do not display gains, 1 = display bin gain in spectrum display, 2 = display long_tone_gain
+	//											 3 = display bin gain multiplied with long_tone_gain
+} NoiseReduction;
+
+NoiseReduction NR;
+
```
db4ple, line 43: A variable does have a memory location and here you can instruct the linker to put it into the __MCHF_SPECIALMEM.
Then it not only compiles but also works as expected. You may get an error from Travis if you have put a little too much into this struct, then carefully move some data over so that you get below the limit.

DD4WH: Perfect! Thanks a lot! Finally I get it into my head :-).
Changed the code and will make another pull request.

PR #1235, opened 2017-11-30
spectral noise reduction bugfix

DD4WH:
* correct allocation of noise reduction variables to the right memory places 
 This should fix the problems with my pull request #1233 

* changed noise reduction parameters for long tone detection = auto notch


df8oe:
Errors inside.... Pse look at Travis lines 1131 and following / 2525 and following.

BTW:
You can get same output locally if you build using Makefile...

db4ple:
Hi Andreas,
also with eclipse you get all the error messages!

db4ple:
But this is an build issue I think, since it only affects the ILI9486 build. The "normal" build works.

df8oe:
Errors of course. But not the same warnings. And some very special situations can take influence due to different building options Eclipse vs. Makefile.

db4ple:
In Eclipse you also get the same warnings if you use the same compiler and settings. Of course, with different options you may get different ones. But here we have probably a simple issue.

DD4WH:
hmm, strange.
I restarted Eclipse and cleaned, refreshed, and rebuilt the index and built the firmware.
No errors, no warnings.
I will try to make the same PR again.

db4ple:
Frank, no. Don't. Does not make sense, the source code remains the same. The travis build does not only build the normal configurations but also the 480x320 ILI9486 display controller one. And only this fails. I am not sure if you can find the problem alone. It is probably something the the code, I will have a look, be patient.

DD4WH:
So, now there is nothing I can do, because I cannot do the pull request, because its already there. Any idea?

df8oe:
Frank, it is not the standard build for ILI932x LCDs. It is the build for ILI9486. You have to define LCD_TYPE=1 via environment variable (don't know this works with Eclipse) or set it manually in uhsdr_board.h. For testing it is neccessary to build minimum:
- for F4
- for F7
- using "old" LCD (LCD_TYPE=2)
- using "new" LCD (LCD_TYPE=1)

I am not sure if its ok only to build both LCD versions for one MCU. Eclipse is not capable to do al that automatically (I think)...

sp9bsl:
Hi Frank,
we use conditional build for new lcd. You can check this with uncommented and  changed 39th line of uhsdr_board.h to:
`#define LCD_TYPE 1`
and then build.

73 Slawek

DD4WH:
I am not convinced. Until yesterday evening everything worked fine without me caring for LCD types and stuff.
So was there a change from yesterday evening in the code from your side?

mchf-eclipse/drivers/audio/audio_nr.h
```diff
@@ -39,7 +39,7 @@ typedef struct __MCHF_SPECIALMEM NoiseReduction
 	//											 3 = display bin gain multiplied with long_tone_gain
 } NoiseReduction;
 
-NoiseReduction NR;
+NoiseReduction __MCHF_SPECIALMEM NR;
```
db4ple: This is the culprit. I did not notice yesterday: You should not define variables in header files, only declare. See
https://stackoverflow.com/questions/1410563/what-is-the-difference-between-a-definition-and-a-declaration
In audio_nr.h you declare:
```
extern +NoiseReduction NR;
```
In audio_nr.c (or any other convenient place) you define:
```
NoiseReduction __MCHF_SPECIALMEM NR;
```
BTW it gets even worse if you define (!) something as "static" (e.g. "static int am_i_here_multiple_times") in a header file. In each compilation unit (aka c file) you would get a different "am_i_here_multiple_times" variable AND no compiler warning. 
The C language isn't an easy language, is it?

db4ple:
Hi Frank, see my comment in the pull request. The problem was in your code, not in the rest of code. I'll explain the background separately to you, it is a C language thing. Oh, yes, you can see my review comment in the history here.

mchf-eclipse/drivers/audio/audio_nr.h
DD4WH: No, it isn´t ;-).

df8oe:
Shall I merge it and fix it here?

PR #1237, opened 2017-11-30
SNR: new VAD by Esch & Vary 2009

DD4WH:
* test of a voice activity detector by Esch & Vary 2009 based on the ratio of the noise-suppressed signal and the noisy signal --> the paper claims this VAD is especially good in low-SNR situations
* use DEBUG menu switch "NR VAD type" to test performance of these two VADs
* Esch & Vary VAD need a considerably lower VAD threshhold --> adjust that accordingly


DG9BFC:
hmmm with the sohn vad i used around 2-3000 as lowest setting ... while esch and vary needs more in the range of 200-300 
what about multiply by 10 and you have a finer setting and both values are in the same value range?!?

default value for long tone thresh is much to low ... it is 200 ... but in my testing that should be much higher (more in the range of 10 or 12 000)
snr gain smoothing?!? i do not find that it does do much ....
it is very difficult to find a setting for nr thresh with vad mode esch and vary ... either the led is on ... or off ...
and reardless what i set it ... sound is similar

PR #1242, opened 2017-11-30
 New noise blanker by Michael now standard NB, old NB deprecated 

DD4WH:
* tried to find all of the many many places, where the noise blanker is treated.
* TODO: could not manage to display the noise blanker setting, at the moment ONLY setable and visible in the menu


db4ple:
Hi,
the NB control via encoder two is if the encoder two is in ENC_TWO_MODE_SIG_PROC. See  UiDriver_DisplayEncoderTwoMode, UiDriver_DisplayNoiseBlanker, and UiDriver_CheckEncoderTwo.

BTW, in case of deactivation for later deletion, clearly mark the areas for instance using an appropriately named #ifdef (best, e.g. __OBSOLETE_NB_CODE__) or explicit comments. This helps to find later all the places to touch much more easily even for you but especially for outsiders it becomes clear what this is for. 

@DF8OE / @DD4WH  I have a pull request (waterfall display change and bugfix waterfall clearing) waiting for the freeing of a tiny bit of CCM (228 bytes) so I need to know how we progress here (I have a broken build on 480x320 @mcHF for that reason).

DD4WH:
Hi Danilo thanks for the help!
fixed the noise blanker display.

PR #1251, opened 2017-12-02
New dynamic spectrum layout

db4ple:
* spectrum layout is now calculated dynamically based on the current hardware and software configuration.
* all display modes and sizes {480x320,320x240} x {BIG,NORMAL size} x { SCOPE, WATERFALL, BOTH} to be available on all resolutions
* waterfall now flows down on all platforms
* frequency display is between scope and waterfall
* it should be now easier to adjust the layout further, such as a half width display.
* there is still a crash issue with x32 zoom in 480x320 and scope enabled.
Some small visual changes need to be done (scope can use a few more lines of height)

Please test!

### DIFF
--- a/mchf-eclipse/drivers/audio/audio_driver.c
+++ b/mchf-eclipse/drivers/audio/audio_driver.c
@@ -2792,6 +2792,7 @@ static void AudioDriver_NotchFilter(int16_t blockSize, float32_t *notchbuffer)
     lms2_outbuf %= ts.dsp_notch_delaybuf_len;
     //
 }
+#ifdef OBSOLETE_NR
 //
 //
 //*----------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


df8oe:
on OVI40 @480x320 now only black area - no scope, no waterfall...

DD4WH:
really awesome display with spectrum AND waterfall on OVI40 & 3.2" display.
BUT: initially no display, I had to press Band+ AND Band- two times until the display appeared.
After shutting down and rebooting: display OK!
Great work!

db4ple:
Hi,
yes. I see, I have to add a check for the 4th state (we use 3 out of 4) and set it to a default. The problem is, our configuration system cannot understand defaults for sets of bits. Will do it programmatically.

df8oe:
Yes, I can confirm it is working great. Maybe a hint for mcHF users would be enough - if not programmatically to solve.

On parallel LCD speed is very good, on SPI LCDs it is much too slow.

I wait for your fix regarding black screen on OVI40 scope/wf area before I merge Franks pull request. I cannot test it without working display ;)

PR #1294, opened 2017-12-17
Touchscreen support for 480x320

sp9bsl:


mchf-eclipse/drivers/ui/ui_configuration.c
```diff
@@ -513,6 +528,17 @@ uint16_t UiConfiguration_WriteConfigEntryData(const ConfigEntryDescriptor* ced_p
     case ConfigEntry_Int16:
         retval = UiWriteSettingEEPROM_Int16(ced_ptr->id,*(int16_t*)ced_ptr->val_ptr,ced_ptr->val_default);
         break;
+    case ConfigEntry_Int32:
```
db4ple, line 531: Hi, I am not a fan of 32bit config variable approach as implemented here.  The reason for not liking it is that we use 2 locations but only reference one in the configuration data description. This calls for trouble. It is now much easier to create "shared use" of configuration variables. This requires careful inspection of every commit in ui_configuration{.c/.h}, i.e. extra work. I couldn't come up with a more clever approach to handle this risk when I refactored the ui_configuration that is why I left this and the reading of arrays out.

Question is: Can we implement a backward compatible approach which is safe to use and evolve and not to complex to use when defining new entries?

sp9bsl:
Hi Danilo,
I was wondering how to do it, all the entries are 16 bit, I needed float or at least int 32bit. I'm ready for ideas :)

sp9bsl:
I reserved two entries in eeprom definition for every 32bit variable (there are 6 of them)

sp9bsl:
I have another fixup for spi speed (badly configured prescaler for SPI in parallel lcd mode, the ts controler worked not properly), can I do the pull request?

db4ple:
Hi Slawek, I would have had a brilliant idea, I would have implemented it.  

When I refactored the code, there were already a number or arrays of 32 bit entries (vfo frequencies) but made in an even worse way ( n x low 16 bit and followed by n x high 16bit, so a single 32bit value was not even using consecutive locations).

I just want to raise the issue that the existing structure does not really support 32bit values in a safe way (or anything else larger than a 16bit value). 
My code to read the original 32bit values is not better in that respect (but I can hide behind the fact that I did not invent the configuration storage approach ;-) ).

db4ple:
Pull Request -> of course, why not? This discussion is not intended to block the use of your code (it is merged anyway).

PR #1321, opened 2018-01-03
Touchscreen. Fixup for ADS7843. Attempt #1.

sp9bsl:
Hi,
I know that this is ugly, prepared it for test just to know if we are going in the right direction ;)

73 Slawek


db4ple:
Hi,
I don't think we should merge this into the official build. It is sufficient to test with this code  and wait with merging until we understand the issue.

sp9bsl:
Ok, agree.

PR #1372, opened 2018-01-28
PSK WIP, experimenting with convergence coefficients #1002

phaethon:


DD4WH:
Hi Eriks,
did a bit of testing:
PSK63 works well! But it is still very hard to finetune, +-2Hz is necessary.
I had a thought, but maybe I am wrong:
You take the audio from the audio path. This is what the audio signal passes BEFORE you take it and decode BPSK:
- I&Q -> lowpass polyphase FIR decimation filter
- Hilbert filter 0/-90degrees
- the sum of I&Q (USB) is taken as the audio signal
- IIR lattice filter 10th order
- IIR biquad for bass, peak filter, manual notch filter
I wonder if the phase of the signal is preserved well enough after all these IIR filters to accurately decode BPSK?
Or would it be better for you to take the raw I&Q signals and then directly do the decoding without any IIR filters before decoding the signal?

phaethon:
Probably, I would need to test it with raw I&Q signals. On the other hand when I use fldigi, it is, also, processing signal from audio path and is still doing better, which shows that there is still room for improvement.
My issue is mainly around the testing. It seems that the simulation on the computer (using audio file recorded on the computer connected via USB to MCHF) with the same psk.c code are doing much better than the one running on MCHF. Is there any filtering/processing between the values that get passed to ProcessSample and the audio forwarded to USB audio interface? If not - has anybody established a way to export the signal for testing outside the device (I mean audio path, I am aware of sending raw I/Q)? I am not sure if there could be important differences because of the nuances in processing floats on ARM and float processing in the way I compile it for PC.

db4ple:
There is a lot of additional processing: the IQ audio send via USB is 48khz and comes straight from the audio codec, no processing whatsoever done. Your code uses iq phase/balance corrected, demodulated SSB audio filtered and downsampled to 8khz. Oh boy, we do processing here...

That may explain the difference. If your code includes your own filtering, it may be a good idea to grab the audio data earlier in the processing queue just like the FreeDV code does. For RTTY it does not seem to make much of a difference. But phase correctness etc. is not relevant for RTTY.  

I don't think the float processing differs much as both FPU should deliver the same result as the operations are based on IEEE... float math standard (I must say, I am not an expert here but assume the standard works this way and does not just define the float number representation itself)

73
Danilo

phaethon:
Not the IQ audio (which I would get putting it into DIQ mode), but "normal" audio from MCHF to PC. Is it processed differently than the audio path internally?

db4ple:
I see. Yes, some processing is done. But it is only upsampled back to 48khz and then straight to USB. So yes, slightly different but it is not much processing.

PR #1374, opened 2018-01-29
LCD driver unification

sp9bsl:


mchf-eclipse/drivers/ui/lcd/ui_lcd_items.h
```diff
@@ -9,14 +9,175 @@
...
+} touchaction_descr_t;
+
+typedef struct
+{
+	const touchaction_descr_t* actions;
+	int32_t size;
+} touchaction_list_descr_t;
+
+typedef struct {
+	uint16_t SizeX;					// Lcd dimension
```
db4ple, line 57: Hi, good stuff. But if you don't mind, I'll go and change X,Y pairs to use a coord data type with x and y being members.
Reduces the number of entries a quite lot. For single X or Y we can still use uint16_t, no big problem. 

But first let the dust settle, this is a simple enough refactoring.

sp9bsl: Hi Danilo,
yes please do it (in fact I was waiting for your refactoring ;) ), there are more like this in this structure. And new ones will follow when we start with 800x480.

sp9bsl: After the today's issue with bootloader I had to add SizeX and SizeY data to  mchf_display_t (MAX_X and MAX_Y)because ts is not available in bootloader code. So we have some kind of redundant data, but for this moment I do not have idea how to (if) solve this.

db4ple: I think we should use the SizeX and SizeY or something along those lines as much as possible and deprecate direct use of MAX_X and MAX_Y whereever possible. And the resolution belongs to the graphics driver and not in the ts structure directly, This ts beast is way too unstructured anyway.

PR #1399, opened 2018-02-01
Added freq display font selection.

sp9bsl:


mchf-eclipse/drivers/ui/menu/ui_menu_internal.h
```diff
@@ -92,6 +92,7 @@ enum
     MENU_METER_COLOUR_DOWN,
     MENU_DBM_DISPLAY,
     MENU_DBM_CALIBRATE,
+	MENU_FREQ_FONT,
```
UR7FM, line 95: How to change  freq font? Im display menu I didnt see that.

sp9bsl: Which version of CPU do you have? This is available for 1024kB+ version.

UR7FM: 512kB :((

sp9bsl: Fonts needs memory. We had to limit unnecessary things (from HAM point of view) to keep support for older machines, sorry.

Issue #1011, opened 2017-08-21
CW Mode: start of first character clipped with external keyer or straight key


ve3ids:
I am having problems trying to use an external keyer or straight key. The internal keyer works great. The start of the character is clipped each time the mcHF goes to TX mode. If the txrx delay is increased, it will send ok as long as the delay is high enough so it doesn't drop between characters. At a word space it will mess up the first letter again. I have a version 6 with the August 19 daily build. 
Thanks
Don ve3ids

la2fda:
Like
https://github.com/df8oe/UHSDR/issues/659 and https://github.com/df8oe/UHSDR/issues/669 ?
I think you find a solution here.
73
Stein

db4ple:
Hello Don,
according to my measurements when working at #669 we had about 23ms delay going from RX to TX in CW with straight key. There may be additional delay before we start processing the first character since we have to wait for the main UI to decide to go to TX. This delay depends a) interrupt load (filtering & DSP use) and the waterfall / scope settings. Please disable waterfall or scope ("increasing" speed to 0 will disable scope or waterfall) and test if this provides improvements.

73
Danilo

ve3ids:
Hi Danilo and Stein,
Thanks for the info. That is the exact problem I am seeing. Disabling the bandscope and waterfall does help. External keying will work up to about 15 WPM now. Above that and it still messes up. I am assuming this is a processor loading issue that the small MCU can't handle? I tried increasing the TX initial mute delay and that made it much worse. I can send slowly with the hand key with the display turned off, it would be nice to have high speed external keying but I understand the hardware limitations. This would be a good feature to consider in the new design. 
73 Don ve3ids

db4ple:
Hi Don,
could you so kind to rename the issue to be a little more descriptive (I cannot do this). I would recommend to "CW Mode: start of first character clipped with external keyer or straight key" just to let the others know what this is about.

The problem is not directly a hardware problem, the SMT32F4 is absolutely capable of responding much faster, it is a problem of the current software architecture. I have an idea how to solve this but it is not trivial to implement so that it works without causing problems.

So stay tuned.
73
Danilo

ve3ids:
Thanks Danilo for your interest. I would like to have a solution to the problem and I realize it will be considerable work. I am really happy with all the other features of the rig and am amazed at the great features you fellows have been able to incorporate into it. I am enjoying it with the fine internal keyer in the meantime. Thank you for all the hard work that has been put into it.
[...]

PR #1404, opened 2018-02-03
autonotch enabled and working

DD4WH:
* autonotch working (LMS algorithm)
* enabled for all machines, even for small mcHF (but beware, could crash, please report)
* no need to adjust parameters for autonotch, but experienced users may do so in the configuration menu

mcHF users, please report if NR plus autonotch works for you.
--> please report for two situations
* filter bandwidth <= 2k7
* filter bandwidth > 2k7 and <= 4k8

*


sp9bsl:
Bad news: it won't build with mcHF ~2.5kB overflowed the ccm memory (i think same for 256kB version because all F4 have same size of CCM). Please look at the Travis output: it says same.

df8oe:
For me, too. Does not build for F4. Actually we have broken binary - I will roll back and wait. Only for Binary - not for GitHub

DD4WH:
Hi Slawek,
thanks for reporting. 
I made a new pull request which puts the data into another part of the RAM. If that does not help, Danilo will have to look at this.
73 Frank

DD4WH:
@sp9bsl @andreas:
I do not have time today any more, so could you please go into that:
In spectrum.c line 374 I introduced 6kbytes of RAM usage :-)
That speeds up the spectrum/waterfall heavily!!!
But it uses too much RAM for the mcHF obiously.
We have to have a 512 array for the small display, and a 1024 array for the large display.
And for the SMALL mcHF we do not assign these arrays at all, but use the commented out code in line 416.
I can do that tomorrow or you do it today ;-).
73 Frank

df8oe:
This is already implemented and it works using your latest RAM approach. I think we only need testers to confirm. Optimization can follow later.

sp9bsl:
Hi Frank,
but this is const array (both of them) so, it goes to Flash not into RAM. Your solution is good - it doesn't consume RAM. If I understood you right: you want to continue with calculation (the commented 416 line) for SMALL and do it by LUT for others?

sp9bsl:
I've seen your addition few days ago and I sit quiet because we have bottleneck in other place: at the last stage of this state machine - when the drawing of spectrum/waterfall occur. This is VERY time consuming process and when I have something like that in code I always do the another "level" of execution: simply an interrupt which has lowest priority and on this new level place all more time critical stuff moved from the bootom most level. On the bottom level (main loop) I leave the most time consuming routines. It needs sync between the levels but usualy helps. Here we have mainly the DMA transfer (SPI mode) and stalls for 10MHz FSMC transactions (parallel LCD). We could optimize the parallel interface also to work in MEM to MEM DMA mode. In both of this interfaces core is "deadly bored" when waits for end of DMA. So we can enable it to do other things in meantime one level above. Just an idea.

DD4WH:
Hi Slawek,
good that it works now for mcHF!
sorry for the RAM/flash confusion, but I am a biologist / non-programmer, so there are many things unknown to me (or things that I need to be reminded of in order to remember again ;-)).
So, OK, we should leave the von-Hann-window-arrays alone (but we could implement switches around them, because with large resolution screens you should always take 1024, and with low resolution screens you should always use the 512 array). The calculation should always be omitted, because it requires 512/1024 cosf-calculations per frame ;-).
Oh, your proposal regarding state machine/interrupts is much too high-level stuff for me . . . sorry, I cannot be of any help there.
Would be good if you discussed that with Danilo!?
All the best 73s,
Frank

sp9bsl:
Hi Frank,
I'm impressed! You're doing really good things and by this you are the programmer!
Regarding the tables: I will look where to put the switches at the evening or tommorow.
I will discuss my idea with Danilo anyway.

73
Slawek

db4ple:
Hi,
I gave offloading the graphics update to DMA interupt a good bit of thought. The main challenge is that all UHSDR code is assuming completely synchronous operation, i.e. it does not contain any sychronization for concurrent access to shared resources such as the graphics memory update via the display controller or the switching from RX to TX. 

Introducing these concepts requires 
* the proper handling of concurrent tasks by means of synchronization 
* better understanding of the related concepts by the programmers of the software 
That means even more carefully programming since concurrent execution introduces a lot more complexity to debugging. I think that would raise the bar for volunteers significantly. Especially on the mcHF were we are short on resources (memory and performance). For that reason I am a bit reluctant to go down this path without consent from the significant contributors. To facilitate this we should do this in a first round in an experimental branch. I am against a half-baked solution hacked together (I do have some experience in the area of design and implementation of embedded RTOS, I have to admit, so I am biased).  

Technically we should base the software on a lightweight RTOS such as FreeRTOS which is not much more than a scheduler and some synchronization primitives with very low overhead and "directly" supported by the STM32 HAL. Of course there are more capable RTOS out there but I have the feeling that we would have to rewrite even larger portions of the code if we go done this route. But I have no experience with these OS, so I may be wrong.

sp9bsl:
Hi Danilo,
it was just an idea. I think I understand the complexity of this solution in case of UHSDR and potential problems for other programers. I think about this quite often but I'm not so brave to start coding it :)  Well, I don't know the RTOS because I'm bare metal programmer. Before we start (if) to do anything in this direction we have to modify the parallel interface way of sending data to lcd (move to DMA mode).

db4ple:
BTW,
what we can do in order to mitigate the specific problem with the display update as being very time consuming and preventing switching to TX for too long in CW modes, we could check after every line if we have a request for TX and simply skip to the end. This is easy and safe as it builds on what we already have. Although this solves only a specific problem, it might be sufficient for now.

sp9bsl:
Yes, this is good idea.

db4ple:
I think moving towards a DMA transfer for memory is not a big deal on the firmware side as it is essentially similar to the SPI DMA. However, it might be too fast for the display controller (not sure, but DMA will probably a little faster then the loop we use now). For the above idea this is not really necessary.

sp9bsl:
No, FSMC will halt DMA due to FIFO full. We have limit ~10MHz there.

sp9bsl:
But of course agree: this migt be not really necessary.

db4ple:
FSMC -> what I meant is the higher memory write rate if we DMA through the FIFO as opposed to doing via individual writes and a number extra instructions in between. May or may not be a problem
Anyway, #1409 implements the quick dropout, curious to see reports for this.

sp9bsl:
Great, I'm not in CW so will not help with testing. Please reference this pull to issue where someone asked about this few days ago.

db4ple:
Of course, just did that. (#1011)

df8oe:
Hi,

I do have some experience with RTOS. Not on actually embedded devices, but in principle. I was involved at RTOS-UH at Uni Hannover and have developed some parts of it in Assembler. I am far away from saying "I do have good skills" - but it is not "a new universe" for me. As I came to mcHF firmware my first thought before I dug into the code was transferring it to a "OS like basement". I like that, I know the advantages  that something els controls timing... That is why I like to work on Linux (not RTOS - but I do not have to care about many things)... As I saw monster functions like they were in ui_driver.c I stuck up with the idea to port it to RTOS. My skills are not good enough for this task. And that was the brake - not the task itself. OK - code is much, much, much better structured than in the beginning - but it is a huge task today, too. I would like to see us going this path - and I agree we should do that in a new branch.

There are some (more or less complicated??) tasks which must be done in parallel. There are already some users have OVI40 carrying H7. So H7 support must come up the next time. Ideally it is running when RF boards are available. So we must discuss about virtual EEPROM on H7 (and I think on F7, too). We do not need this relict. We can fit SPI Flash, or use sd card which we can access via SPI.

@Danilo: Porting to RTOS would be a task that is mainly your part. This would be something I am very interested in and will try to follow - but I have much time to spend in OVI40 hardware. I already have a lack of time - sadly...

DD4WH:
Hmm, maybe my comment is completely off-topic, because I do not know too much of the things you are discussing, but my impression is the following:
* at the moment we have only one bottleneck (which is in spectrum.c)
* that bottleneck could be easily solved, if we divide the drawing of spectrum/waterfall into several parts
* another solution would be the RTOS, which would take many many hours/days/weeks of implementation: however that would solve similar potentially coming problems also

--> to me it seems it is not really worth the effort, because:
* a solution for the known existing bottleneck could be more easily implemented by other means
* the RTOS implementation would be very time-consuming
* other tasks would -from my point of view- be much more interesting and "urgent" to implement: for example: 

WISHLIST:
- VOX
- a nice TEXT-WINDOW for sending and receiving text for CW & other DIGIMODES
- the new FREEDV-mode
- maybe also an SSB-finetuning mechanism
- ECSSB (envelope controlled SSB)
- predistortion for the power amp to improve IMD
- etc.

What do you think? :-)

73 Frank

PR #1405, opened 2018-02-03
trial to fix memory problem in mcHF autonotch

DD4WH:


mchf-eclipse/drivers/audio/audio_driver.c
```diff
@@ -590,7 +590,8 @@ LMSData            __MCHF_SPECIALMEM lmsData;
 #endif
 
 #ifdef USE_LMS_AUTONOTCH
-LMSData            __MCHF_SPECIALMEM lmsData;
+//LMSData            __MCHF_SPECIALMEM lmsData;
+LMSData            lmsData;
```
db4ple, line 594: The CCM memory aka MCHF_SPECIALMEM is very much full, adding anything here may cause problem when linking (linker reports overflow). This is not visible on non-mcHF builds as these don't have this kind of extra memory and use the "normal" RAM here as well.

That report can be seen from the Travis build report which I recommend to turn on for the own repository or to at least checkout the result of the Travis build as shown in the pull request (the green checkmark or the read error cross, appears once the Travis build finished)

DD4WH: OK, understood, thanks for the important hint, Danilo!
Will check the Travis report next time I fiddle around with mcHF memory.
My original thought was: I took some variables out of CCM, so I tried to put something else in. But obviously I put more in than I took out before. :-)
73 Frank

PR #1407, opened 2018-02-03
Simple refactoring of scope window function.

sp9bsl:
This was discussed with Frank DD4WH.


mchf-eclipse/drivers/ui/lcd/ui_spectrum.c
```diff
@@ -365,12 +365,13 @@ static void UiSpectrum_UpdateSpectrumPixelParameters()
...
+#ifndef IS_SMALL_BUILD
+// Information on these windowing functions may be found on the internet - check the Wikipedia article "Window Function"
```
db4ple, line 369: Hi,
please DO NOT USE IS_SMALL_BUILD in this way. It is not meant to be directly referenced inside the code. See my comment right were it is defined.
Please identify the variation you do here with a name and introduce a "USE_<NAME>" define (please document in uhsdr_board.h(.
 
Then add the USE to #ifdef IS_SMALL_BUILD ... in uhsdr_board.h

Rationale behind this:
It will allow us to individually add or remove some of these variations to a small build. More flexibility, as if we spread IS_SMALL_BUILD everywhere in the code which will at some point into a mess.

I probably should add these instructions to the comments already there AND in the wiki...

sp9bsl:
Hi Danilo,
ok, agree. I will create a USE_ with apropriate name.

Issue #1417, opened 2018-02-05
DSP A-Notch 


ly2bok:
V. 2.7.92 . SDR mcHF v.04 . Mem. 512.  DSP A-NOTCH  not working perfectly. Fully distorts the signal and makes a noise like a steam locomotive.
http://www.qrz.lt/ly2bok/Laikinas/A%20Noch.mp3
LY2BOK Justas

df8oe:
Cannot confirm. Works as expected on OVI40 and mcHF (STM32F429@2M].

Any other confirmations?

ly2bok:
How it sounds, so I wrote it down. The interference is suppressed well, but this distortion is also present on the NR, and on the A-NOTCH, NR + HOC
the sound  of the "locomotive" is added.

df8oe:
I cannot hear this here. Not at OVI40 and not at mcHF. What LCD and LCD mode do you have?

ly2bok:
HY28B   SPI  Controler  ILI9325

df8oe:
That may be the problem. SPI driven LCDs are much more noisy than parallel driven. I do own one mcHF using SPI LCD and there the noise is slightly to hear, too. The sound is known as "helicopter-noise". Best will be to drive LCD in parallel mode, I think... Not a firmware issue.

ly2bok:
Toutcscreen will work in the   LCD the parallel mode?

df8oe:
Yes, of course. LCD and touchscreen do have different SPI lines and can ba used independently.

ly2bok:
OK I'll do it.Thank.

Harry-ei6ku:
I too have this low level helecopter noise evident on my mcHF, however on my radio it can be eliminated by setting NB to zero. (Its a clone radio, ver 0.4 board with 512K ) .

df8oe:
If you set NR to ZERO you switch it OFF. NR identifies this noise as "wanted signal" and craps it out of the noise floor.

Do you know if your LCD is in SPI or parallel mode? I am not sure but I think clones are working in SPI because they do have RTC... And they are "obsolete at buying". It is very bad to fit a 512KB MCU where it is known since early 2015 (!!!!) that 1M MCU is recommended...

Harry-ei6ku:
[...]

PR #1420, opened 2018-02-05
disable switch for small mcHF to not use predefined window data for s…

DD4WH:
…pectrum/waterfall display

There is an issue report of an mcHF producing stuttering audio #1417 . Maybe this can heal it!???
Should save considerable amount of processor load during spectrum/waterfall calculations


sp9bsl:
Hi Frank,
you may disable it in uhsdr_board.h.

DD4WH:
OK, thanks Slawek, will change that :-)

df8oe:
I suggest that does not change anything for the reported noise - but reality will show :) I wait for correction PR.

DD4WH:
Hi Slawek, hi Andreas,
you are absolutely right: the small mcHF already uses the const windows!
I was confused by the "#ifndef" vs "#ifdef" :-)
So I close this pull request!
73 Frank

sp9bsl:
USE_PREDEFINED_WINDOW_DATA is not used for SMALL_BUILD.
If you want to enable it for all types, simply move it outside the "#ifndef IS_SMALL_BUILD"
Yes those "#ifndefs" sometimes are misleading :-)

sp9bsl:
If there is not enough horsepower available they should stop drawing of scope/wfall by simply go to MENU.

PR #1421, opened 2018-02-05
Fixed out-of-bounds error which causes firmware crash

db4ple:
The wrongly defined filter state size caused crashes in older gcc compiles with -O2
and probably unknown other problems if the firmware did not crash
This fixes -O2 builds for gcc 6.3.1 (and potentially others)
@df8oe: Please set -O2 in Makefile for F7 builds after confirming my fix works.
@DD4WH : Please check the buffer size definitions of the FIR & IIR Filters in audio_filter.c, I am not able to judge easily what is right and wrong here was just able to see that arm_fir_init_f32 used different value then size of supplied state buffer. Yes, C is a very studid language in that respect.


DD4WH:
Hi Danilo,
fantastic, that you found that error!
I will go through all the FIRs and IIRs (not only in audio_filter.c), including all the decimation and interpolation filters and check them.
I will also try to change the nomenclature of those #defined constants, because the names are misleading sometimes.
73 Frank

Issue #1319, opened 2017-12-31
Option to flip the menu navigation for side-mounted encoders?


frazierjason:
The mcHF v0.7 board moves the general purpose encoders to the side of the unit, which presents a point-of-view issue for navigating the menus and changing values.  On all previous UHSDR builds, these encoders are oriented with the shaft pointing at the user, which turning clockwise would increase a numeric value, or navigate "down" in a menu.  But with the shaft parallel to the user's torso and pointed toward the right, the encoders now have an experience of rotating clockwise ("up") to increment values which makes sense, but scrolling in the menus it instead performs a down navigation event.  Can this be fixed, and if so, is it best done in a mcHF v0.7 specific build, or would a new configuration option be the best way?

df8oe:
Yes - this question is very interesting. New orientation of encoders do make different directions mandatory. Because of hardware of mcHF 0.7 and former versions is not firmware detectable (all the same...) the best would be a menu selectable way. We are not sure if it will be an "one config entry point" (like "mchf-version") or many config entry points (like "enc1 dir", "enc2 dir"). But this is on our todo list.

vy 73
Andreas

frazierjason:
I noticed one difference in the v0.7 schematic that was not present in v0.6:  a 4.7k pullup resistor on I/O pin PB2 (BOOT0) that goes to the BAND+ button situated across the anti-spark capacitor.  Were other versions prior to v0.6 also lacking that 4.7k pullup resistor?

I wonder if we could do any kind of trick to detect that pullup resistor, which would tell us we're on a v0.7.

df8oe:
This can be used to identify 0.7 boards. During boot we can add a pulldown resistor to BOOT0 and check if level at PB2 is low. If it is, we are on older revisions, if not, we are on 0.7. Possible issue can be that external resistor is to high so that level does not increase enough - but simple replacing (e.g. with 1K) would solve that.

df8oe:
But again I do not completely understand what is the issue (I do not have access to 0.7). Should directions of all encoders swapped at 0.7 boards? Or only some functions (would be much more different)?

frazierjason:
The encoder direction swap only needs to occur for v0.7 on any encoder action that visually correlates to movement of a visually vertical UI element.

For Enc2 this means logic reversal in UiMenu_RenderChangeItem(), to ensure menu item selection travels up and down corresponding to the twisting of the Enc2 on the right hand side.  

[...]

PR #1433, opened 2018-02-09
User-selectable menu inverse scrolling for Enc2/Enc3 encoders

frazierjason:
Add display menu option to invert the menu navigation for Enc2/Enc3.  This is intended to use on radios with encoders mounted on the right hand side (such as mcHF v0.7) to allow a more intuitive UI experience.  When turned on, turning Enc2 clockwise will advance the selected menu item vertically up toward the top of the screen, and turning Enc3 clockwise while focused on an opened menu group will hide the group.

No other encoder operations are affected; turning the dials clockwise will increment/advance all other selected values as usual.


frazierjason:
Forgot to mention, tested and works great for me. The navigation changes worked as expected for the last two weeks, and I tested the save option lay night on v0.6 and v0.7 units.

sp9bsl:
Hi,
it works, but wouldn't it be better to have reverse option on all encoders saparately  instead of this?

db4ple:
Hi,
we should not mix things here. The pull request does not makes it possible change the encoder direction in general, it changes the direction the menu code interprets the encoder direction. That means as the pull request message says (Thanks Jason, this one was very good)., encoder direction does not change in general. And cannot be solved by implementing a change of encoder directions per encoder (as this would not differentiate between Menu vs. not in Menu operation).

BTW, I am not saying that I think that the changed direction makes sense for me personally since I never used a 0.7. I would rather think that the normal direction would be the "right one" for me. But I would have to try this on the real thing to be sure. 

I would also suggest not to put this in the Display menu but in the Configration menu (which is intended for mostly one-time configuration items) menu, but that is a minor thing.

frazierjason:
Hi Danilo,

Thanks for sharing this info on the menu group intents.  I am not opposed to someone moving the menu option from Display to Configuration menu group, if there are strong organizational feelings about where to put it.  But this new option addresses a cosmetic display-related visual issue, so I figured people would look in there since it feels like the screen is going the wrong way when it's not set up right.  If you are correct that some people will feel the non-inverted experience is the "right one", then this becomes not a one-time configuration, but something that might get reconfigured for personal taste every time the radio changes hands.

If there are no other objections, then we can close issue #1319.

73 de Jason W7DM

Issue #1131, opened 2017-10-15
Keyer memories improvements


ve3ids:
The 3 memories that are implemented are working well. Storing them to the EPROM to save during power down would be great. Also, having a repeat function for the memories with a programmable pause included would be a good benefit as well for contest operations.  (Beacon)
Thanks for the hard work. 
73 Don ve3ids

9a6cw:
finally excellent feature !! thank you very much ! It works ! two memories for different cq are enough..
73 Ivica / 9a6cw

S52CQ:
Hi, I am testig CW memory keyer. It does work great, 3 memories are more than enough. Is that a development feature that memories are not stored in the EEPROM?
73 de Jure, S52CQ

db4ple:
Yes, for now we do not implement the EEPROM but we will.

s53dz:
An excellent job done with CW memories implementation and straight forward procedures to use them.

TNX 73 Bojan

S52CQ:
During WW contest last weekend I was using mcHF CW keyer memories with great success however I have found a bug. 
When in CW mode all works well, although it's rather annoying that after keying is over TX still "hangs" for 100 or 200 mseconds, while in a contest one would require immediate switch to RX. This is not the bug - that is an observation.

Real bug is this: set mode to LSB, press play of any of memory buttons. LCD would display > sign, but nothing happens (as supposed). You can do that few times and few > signs will show. But if you then switch to CW and press that memory button all previously pressed memories are transmitted - without possibility to brake it.

This has also shown that it would be great to have one button to immediately stop transmitting memory. That would be handy in case that one presses two times same button unintentionally. Trust me - unintentional press does happen especially under 48 hours contest...

As a matter of fact there is one button still free isn't it?

73 de Jure

phaethon:
@s52cq sending PR in a moment fixing the issue with mode change and sending previously sent memories.

phaethon:
@S52CQ button is reserved for potential use. Removing those 100-200 msec after transmitting might not be trivial change in a code. I would like to postpone for later.
I suggest renaming the issue into "Saving keyer memories into EEPROM" to make it more specific. The rest has been addressed in one way or other.

phaethon:
Since today, memories are saved to EEPROM.

S52CQ:
[...]

Issue #236, opened 2016-04-08
Memory Management


HB9LFU:
Hello mcHF programmers

has anybody thought of implementing memory banks for the mcHF rig?

Could they be managed via an external usb keyboard?

vy73s
Roger

df8oe:
Yes - this is already on our "2do-list". This idea was the reason to add serial EEPROM. I want to dd memory banks where you can name memories, import memories via USB-stick & textfile and so on. Will be implemented in future.

vy 73
Andreas

HB9LFU:
Hello Andreas,

then it was a good idea to use serial EEPROMs in my mcHFs. Sounds very promising! Thanks!

vy73s
Roger

db4ple:
Getting support for anything externally connected to the USB is a problem right now, as we do not seem to be able to get the "big" USB connector working software wise. An alternative would be the use of a touch screen keyboard. Not as convenient but at least better under our control right know (but there is no touch keyboard yet and nothing related to memories to be honest, except for the some rather dead code).

HB9LFU:
Does that mean that the "big" USB connector is useless and cannot be activated for anything helpful?

vy73s
Roger

db4ple:
At the momement yes. But this is as far as we can see, it is a software only issue. So there is hope.
Both connectors are connected to different USB controllers inside the STM32F4. In theory, it is very easy to switch but in practice, it does not work yet. Once solved, many interesting options will be available how to use this second USB port.

HB9LFU:
Ok, now that sounds more optimistic!

Maybe one can find information on addressing the usb controllers from other projects using the same processor ! ?

df8oe:
No chance. I have dug the web many hours but no code example which uses the HS USB core in FS mode (thats the mode we need)was to find :(

vy73
Andreas

HB9LFU:
Hello Andreas,
on page 100 and following there is information about these controllers in different modes:
http://www.st.com/st-web-ui/static/active/jp/resource/technical/document/user_manual/CD00289278.pdf

vy73s
Roger

df8oe:
We know this and have acted like this but does not work.

vy 73
Andreas

HB9LFU:
Hello Andreas,

I am sorry for that! When I find details anywhere I will let you have a look at it. 

vy73s
Roger

df8oe:
[...]

PR #1435, opened 2018-02-10
#1131 First try to implement storing keyer memories into serial EEPROM

phaethon:
This is work in progress. It currently works only for the first memory, but I haven't found the bug yet. May be I am misunderstanding how SerialEEPROM_24Cxx_WriteBulk should be called?

Serial EEPROM is mandatory for this f-ty.

Please, comment on the value of EEPROM_KEYER_MEMORY_ADDRESS 
I am not checking the size of EEPROM as the minimum supported size is enough.

### DIFF
--- a/mchf-eclipse/drivers/ui/ui_configuration.c
+++ b/mchf-eclipse/drivers/ui/ui_configuration.c
@@ -577,6 +577,31 @@ static void __attribute__ ((noinline)) UiConfiguration_ReadConfigEntries()
     }
 }
 
+void UiConfiguration_UpdateMacroCap(void)
+{
+	int c;
+	uint8_t* pmacro;
+
+	for (int i = 0; i < KEYER_BUTTONS; i++)
+	{
+		if (ts.keyer_mode.macro[i] != '\0')
+		{
+			// Make button label from start of the macro
+			pmacro = (uint8_t *)ts.keyer_mode.macro[i];
+			c = 0;
+			while(*pmacro != ' ' && *pmacro != '\0' && c < KEYER_CAP_LEN)
+			{
+				ts.keyer_mode.cap[i][c++] = *pmacro++;
+			}
+			ts.keyer_mode.cap[i][c] = '\0';
+		}
+		else
+		{
+			strcpy((char *) ts.keyer_mode.cap[i], "BTN");
+		}
+
+	}
+}
 
 //
 //*----------------------------------------------------------------------------
@@ -653,6 +678,9 @@ void UiConfiguration_LoadEepromValues(void)
     UiReadSettingEEPROM_UInt32( EEPROM_XVERTER_OFFSET_HIGH,EEPROM_XVERTER_OFFSET_LOW,&ts.xverter_offset,0,0,XVERTER_OFFSET_MAX);
 
     UiReadSettingEEPROM_Filter();
+
+    ConfigStorage_CopySerial2Array(EEPROM_KEYER_MEMORY_ADDRESS, (uint8_t *)ts.keyer_mode.macro, sizeof(ts.keyer_mode.macro));
+    UiConfiguration_UpdateMacroCap();
 
     // post configuration loading actions below
     df.tuning_step  = tune_steps[df.selected_idx];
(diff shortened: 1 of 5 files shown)


db4ple:
Hi Andreas, Eriks,
I propose not to merge this now. Let us have a look at it and discuss a little how we are going to use the EEPROM in general before we integrate this.  

@Eriks: I will have a look at your code this weekend and see if I find your problem with storing data.

phaethon:
I am fine with having discussion before merging as I agree that some overall plan for EEPROM usage would be good. Are there any other ideas what to store in EEPROM besides configuration variables, keyer memories, and #236? 
One more idea is simple logging. Since last week I have RTC, may be at some point of time I could look into that.
For voice memories currently recommended sizes are too small.

df8oe:
Hi Eriks,

storing of SSTV images, voice, logging and so on in is planned for SDcard. But this is not mcHF related - only OVI40. Keyer memories CAN be stored there - but it will be very wise to discuss the structure of the storage.

db4ple:
Hi,
In order to combine flexibility with efficient use of available storage, I suggest to consider a block-wise chained memory allocation scheme with flexible data length. I.e. a fixed size block consists of a header (id, len) and storage space of a fixed size. A storage space can span multiple blocks.  Think of it as a very simple file system, where each id represents a "file". 
We could use this to store arbitrary large data structures such as strings, memories, etc. Different from a file system I would suggest that once a "file" has been written for the first time, it is not allowed to grow, only to shrink or to be deleted. 
This makes management relatively easy.

phaethon:
Essentially this is implementing malloc. Do we really need it right now? It might make sense for dynamically allocating/freeing memory for received SSTV images, but for configuration information, which is static, it should be sufficient just deciding on the address and consider it reserved for other uses.
I am just afraid such significant complication as it would postpone this PR, which solves something that people could use right away. Then we can later refactor this if needed.

df8oe:
Hi Eriks,

we are not lucky to burn EEPROM cells with structures which can (and will) get problems if structures change later. There are already wishes for storing more informations to band specific storage places, and there will come up frequency memories which will need all the storage of a normal band specific storage place. I feel that a "data structure" like Danilo proposes will  work for all extensions that will come up the next future.

Additinally we should pay attention for maximum capacity of virtual EEPROM. There will be overflow in the future, so no backup possibility from this point on.

And the last, most important thing: first all memories should work not only one.

Let us discuss about a file system like data block system and define the first identifiers.

DG9BFC:
just an idea ... memory for keyer should also be usable in other modes ... and editable with keyboard
so then we have "macro" like texts that can be used in cw or rtty or psk (or any othere mode in future)
i would guess that the keyer (or better say the memory of to be send text) could be useful in any mode
(cq cq de callsign pse k can be used also in psk mode or rtty as an example)

phaethon:
@DG9BFC Memory keyer is usable for playing recorded message in PSK from the start.

DG9BFC:
ok on that ... will the memory texts be available in ALL modes (rtty included)? then they also can be used in whatever mode will come ... and that was the idea behind my suggestion (available in all modes) ... GOOD WORK, WELL DONE

db4ple:
@phaethon: I don't see that we get any progress on the general eeprom storage code. I  suggest to change the start address to 0x1000h . Minimum accept EEPROM size is 32k, so 4k boundary is okay, I'd say.

phaethon:
Changed memory start address to 0x1000. Now it works for all 3 memories.
May be we can reserve part of the memory for "static use" (like in my code), and other part for advanced memory management (to be implemented later on)? As I wrote earlier I don't see the immediate benefit of advanced mechanism right now (at least not yet and not for such use cases where fixed amount of memory is needed as for this PR).

db4ple:
Hi Eriks,
unfortunately it seems to me you did not use `git rebase` as described in the contribution guide. You probably used `git pull` from your github repository, and because of that your pull request now contains 40 changed files coming from 56 commits. We will not merge the PR in this form. Sorry. 
But I think you can quickly resolve that by running a `git pull ...` followed by `git push -f ...` (please see instructions for details). If you need help, let me know. 
 
Thank you!

EDIT: To be honest: Maybe these extra commits will go away once merged by Andreas, DF8OE via git, but we won't that risk here.

phaethon:
@db4ple please, see now.

db4ple:
Thanks Eriks, result looks perfect.

Just for my curiosity did it work more or less with the approach I suggested?

phaethon:
Roughly, yes. In addition, I did `git reset ...`

PR #1446, opened 2018-02-13
trying to work this out still guys

quadrant2005:
fingers crossed this will submit to you guys...  Ive been ill the last few days so not been about...


db4ple:
Hi Rich,
this pull request is for sure not what you want us to integrate. Instead of submitting changes to the mchf-github project you submitted your Eclipse workspace (which also includes the project but a lot of stuff which should be committed).

It is not all your fault, but still not acceptable. In general it is not a wise idea to simply add and commit all files but here this went really wrong. Please remove  folder .metadata and all content from the github repository. This data is Eclipse internal data which cannot be shared.

quadrant2005:
Thanks i will try again without eclipse things sorry about that, that explains why it took so long to upload, Not very well at the moment but wanting to try and accomplish something

Issue #1450, opened 2018-02-14
Treble control stops audio (BUG!!)


DG9BFC:
Your firmware version: 2.9.4
Your bootloader version: 2.0.0 ... but does not matter here

### Hardware 
 * UI Board: mcHF 0.4
 * RF Board: mcHF 0.4

### Describe the issue:
the treble control stops audio when turning down below zero (exactly at -2)

You have audio related problems:
 * DEMODULATION MODE: LSB
 * Filter:  issue is in all filter settings
 * Noise blanker setting: 0 (regardless if NB is on or off)
 * Noise reduction: ON/OFF (regardless if NR is on or off)

You have display (also fill if audio problem) related problems:
 * Display resolution: 320x240 (old 2.8 display)
 * Display connected via: Parallel 
 * Touchscreen: Yes
 * Magnify: 1x 
 * Spectrum: Scope

the treble control does not work like in older version

2.9.3 is ok ... 2.9.4 is not

sp9bsl:
I can confirm this on mcHF 1MB/192k wit 2.9.4. With BASS=TREBLE=0 all is ok,

db4ple:
Hi,
thanks for this report, I think this will allow me to fix the issue quickly. Good job!

db4ple:
Hi,
if you change bass or set treble to a positive value -> does this work?
if you save -2 and restart, there is no audio, with -1 there is? Correct?

DG9BFC:
when saved with -2 after power on a very short noise burst ... then audio stops
when saved with -1 audio works
that is treble
bass setting is exactly reversed
means minus settings in bass do work
plus setting breaks audio (+1 works +2 breaks)

db4ple:
Hi,
I could not reproduce the issue on 2.9.4 on the OVI40 so please (once Andreas merged my pull request) check out if this fixes the issue on the mcHF. Too be honest I have no idea why it is -2 or +2 which causes the problem. It is difficult to debug if you don't have machine exhibiting the problem... but my fix fixes a problem we had in the code and it may be the root cause for this issue (or not).

db4ple:
Ok, according to @df8oe issue is fixed by PR #1451. We may have close a longstanding source of interesting problems (without knowing what problems we heard about might have been caused by this issue).
Waiting for additional confirmation from original reporter (Sigi), then please close this issue.
Thanks again to Sigi for finding a way to reproduce the issue and putting me on the right track.

DG9BFC:
[...]

PR #1451, opened 2018-02-14
Treble crashed audio #1450

db4ple:
(Hopefully) fixed audio lost in 2.9.4 if treble is changed to -2
In any case, used array size was too small.

### DIFF
--- a/mchf-eclipse/drivers/audio/audio_driver.c
+++ b/mchf-eclipse/drivers/audio/audio_driver.c
@@ -937,7 +937,7 @@ void AudioDriver_SetRxTxAudioProcessingAudioFilters(uint8_t dmod_mode)
 
     // the notch filter is in biquad 1 and works at the decimated sample rate FSdec
 
-    float32_t coeffs[5];
+    float32_t coeffs[6];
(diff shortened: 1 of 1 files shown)


DD4WH:
Hi Danilo,
an IIR biquad has five coefficients, not six. Has been working for years now in the code with 5 coefficients.
So, I do not believe the issue is related to array size, but to the change that you introduced using pointers.
Just my quick thought.
73 Frank

DD4WH:
one other thing: AudioDriver_SetBiquadCoeffs has only five coeffs to set, not six.

db4ple:
Yes, the filter has only 5 coeffs but during on the fly calculation we use 6 floats in that array (index A0 is 5), see for instance void AudioDriver_CalcLowShelf(float32_t coeffs[6], float32_t f0, float32_t S, float32_t gain, float32_t FS) or AudioDriver_CalcNotch(float32_t coeffs[6], float32_t f0, float32_t BW, float32_t FS), ...

df8oe:
I can confirm the issue is gone now on all BAS/TRB combinations on mcHF/OVI40.

PR #1454, opened 2018-02-15
Fix for current menu item position on menu page change.

sp9bsl:


mchf-eclipse/drivers/ui/menu/ui_menu_internal.c
```diff
@@ -746,12 +746,14 @@ void UiMenu_RenderLastScreen()
         UiMenu_DisplayMoveSlotsForward(ts.Layout->MENUSIZE);
     }
     UiMenu_RenderMenu(MENU_RENDER_ONLY);
+    UiMenu_MoveCursor(ts.menu_item%ts.Layout->MENUSIZE);	//Update menu item to show the current line marker
```
db4ple: Hi Slawek, 

which problem are you trying to solve here? As the original author of that code I have a certain interest in understanding what the issue is ;-)

If the addressed problem is a problem (which I assume it to be), then my comments below apply:

I think we should better move the modulo operation into the UiMenu_MoveCursor  and apply this all the time, no harm will come from that but much good.

And why not put the update of the cursor inside the UiMenu_RenderMenu in the part where we handle MENU_RENDER_ONLY?

sp9bsl: Hi Danilo,
Try to press  PREV or NEXT when in menu. The cursor is not refreshed  (the "<" dissapears).
Render menu - as I see from code - is called with different parameters (render only or change), and to avoid refreshing every time when the item is changed I added it just to those four routines.

db4ple: Hm. Will test tomorrow. Should have noted that one, to be honest. 
I suggest to put the MoveCursor in the case MENU_RENDER_ONLY part, since this is intended to draw a full menu page, and, it makes sense to display the cursor then, too. I briefly checked the other location where UiMenu_RenderMenu(MENU_RENDER_ONLY) is called, and seems save to me to always redraw the cursor when rendering the page. But I did not test it.

sp9bsl: When you call UiMenu_RenderMenu with MENU_RENDER_ONLY you do not update the cursor because at the and of UiMenuUpdateMenuEntry there is check exactly for MENU_PROCESS_VALUE_CHANGE which is different from MENU_RENDER_ONLY ...

sp9bsl: I played around with current item highlight because 14 lines of menu at new 3.5" lcd is quite dense and found this...

Will try to put it inside the MENU_RENDER_ONLY part.

db4ple: Hi Slawek,
maybe we should consider to use a larger font for the menu entries on the 480x320 display. 14 lines are not very easy to keep control off, especially if they are very small in size (which they are).

sp9bsl: Will check if we have other fonts which are a bit larger, if not - will prepare something new.

db4ple:
Hi Slawek,
thanks for changing the commit  More elegant.

sp9bsl:
I just wanted to close this ;) OK, so can you confirm this?

db4ple:
Yes, Andreas should merge this at next opportunity

PR #1455, opened 2018-02-16
Lightweight High Priority Tasks

db4ple:
_Please wait with merging until we have confirmation that #1453 works as expected._

This commit provides a way to run certain (longer running) functions
with a priority above the normal application code responsible
for controlling the UHSDR UI and functions. Examples for these functions
are longer running audio processing functions which need to get priority
before things like the display redraw otherwise audio may sound
distorted (FreeDV and the new noise reduction), we recently had issues like
that on the mcHF / STM32F4.

Technically we run higher priority task by means of a low priority
interrupt handler for the PendSV interrupt (which is kind of made
for this sort of use). All hardware interrupts have higher priority but
normal application code only resumes after our PendSV high priority code
finishes.

Word of Warning: Do not randomly add code to the High Priority task execution.
This code needs to be treated the way we treat all interrupt handlers,
we need proper synchronization via interrupt-safe mechanisms such as
special queue/buffer implementations. Direct calling of UI functions is
not permitted, the code must not run for too long/too often, otherwise the UI will no
longer be responsive etc.

In case of issues, you can turn off these mechanism by not defining
USE_PENDSV_FOR_HIGHPRIO_TASKS in uhsdr_board.h . This is only a fallback,
and will be removed once we have sufficient assurance all goes according to plan.


df8oe:
From my side all is working. How long shall we wait until merging?

db4ple:
Go ahead, I just wanted to make sure that someone was trying my filter refactorings before moving on.

Issue #1470, opened 2018-03-03
NR-NOTCH problem. Slows everything down.


WM0I:
## Your Issue Data goes here:

Your firmware version:  2.9.12
Your bootloader version: 4.0.0

### Hardware 
 * UI Board: mcHF 0.x / OVI40 1.x    Clone
 * RF Board: mcHF 0.x / OVI40 1.x    Clone

### Describe the issue:

Place tell us what does not work and how others can reproduce the issue (list related settings, steps to do, etc.):    When pushing the DSP button to get to the PEAK (755 value) I have to go through the NR-NOTCH setting. This is where the problem comes in. Every thing slows down on the scope to a crawl. all the buttons and all the knobs are not responsive. I have to keep tapping on the DSP key to get it off the NR-NOTCH setting so that I can get back to normal scope and button response. 

I think what is happening is the CPU is not fast enough to handle all the buffers to do the new Noise Reduction and Notching.

Need away to get out of the NR-NOTCH function with having to tap the DSP button so many times to get to normal.

Best regards
WM0I

df8oe:
I can confirm this impacts only radios with SPI LCD. Activating DSP NR slows radio down a little bit, but activating AutoNotch, too makes it nearly unoperable. Working with manual notch does work.

Radios with parallel LCD are NOT AFFECTED - OVI40 not, too.

Question:
If not solveable via firmware because SPI LCD is the bottleneck: shall we disable AutoNotch together with DSP NR when running on a radio with SPI LCD?

vy 73
Andreas

db4ple:
Hi,
that is good and bad news together: It shows that the High Priority Task concept delivers uninterrupted audio even under very high loads (since there was no report about choppy audio) but of course, it cannot do wonders if the overall performance is not sufficient. And the result of this is sluggish response of the UI.

Disabling Auto Notch only for SPI / STM32F4 is technically possible but of course calls for trouble in terms of support requests ("Why can I don't get AutoNotch plus NR to work?") even if we mention that specific behavior in the documentation. If there is no other solution for now, we should go this way. 
Let us think for a brief moment of other ideas. If nothing turns up by the next days, lets do the disabling of the combination.

DD4WH:
there are two things we can/should do:

1.) alter the way the DSP box works. Let us think about a way to choose the different options without having to toggle through all the modes. Also some combinations of DSP goodies are not yet available at all (for example peak + NR, or manual notch + NR which could be very useful!)
[...]

PR #1471, opened 2018-03-04
Interim fix for #1470 (mcHF slowdown if autonotch & NR)

db4ple:
this code is a hack to reduce processor load for STM32F4 && SPI display
which causes UI lag
in this case we simply use a less power-eating filter (lower number of taps)
one problem is that we use the not so good filter
even if the autonotch / nr is not active and we could use the good filter


DG9BFC:
will that influence all mchf or just when spi is used??

db4ple:
As it says filter is used for mcHF AND SPI display only, all others get the better one. It is decided at runtime so that we can use the same binary for all mcHF.

PR #1475, opened 2018-03-15
potential AGC bugfix

DD4WH:
maybe this helps in fixing AGC bug when receiving in AM/SAM with filters >4k8
my hardware is on repair at the moment, so this is an untested pull request!

Please test with demodulation mode AM/SAM and filters lower and higher than 4k8 before merging!

### DIFF
--- a/mchf-eclipse/drivers/audio/audio_driver.c
+++ b/mchf-eclipse/drivers/audio/audio_driver.c
@@ -746,7 +746,7 @@ void AudioDriver_SetSamPllParameters()
 
     // definitions and intializations for synchronous AM demodulation = SAM
     //    adb.DF = 1.0; //ads.decimation_rate;
-    adb.DF = ads.decimation_rate;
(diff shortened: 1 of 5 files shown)


mchf-eclipse/drivers/audio/audio_driver.c
```diff
@@ -746,7 +746,7 @@ void AudioDriver_SetSamPllParameters()
 
     // definitions and intializations for synchronous AM demodulation = SAM
     //    adb.DF = 1.0; //ads.decimation_rate;
-    adb.DF = ads.decimation_rate;
+    adb.DF = (float32_t)(ads.decimation_rate);
```
db4ple, line 749: Cast not needed. Automatically converted during assignment.

df8oe:
I cannot test because I never have seen the bug which is fixed with that. So I merge it and let do the test the ones who observed it.

Issue #1438, opened 2018-02-11
In AM mode AGC


ly2bok:
In field versions тhe AGC in AM modulation does not work. If the level of the signal is applied to the receiver input from the generator (-80 dB), and then the signal is turned off then the receiver does not return to normal reception. It remains locked at that level and there is no Rx  sensitivity. Even when changing to another modulation, the AGC "does not let go." Only "Power" helps.
Justas LY2BOK

DD4WH:
Hi Justas,
I listen a lot in AM and SAM and never had that problem. I just checked and cannot confirm.
Please let us know some more details:
* hardware type & display
* filter bandwidth
* NB setting ?
* NR setting ?
* autonotch ?
* AGC settings: for AM/SAM, I highly recommend the standard setting: vLO, which means a very slow AGC!
* what is your AGC threshold?

DD4WH:
1 | 2 | 3 | 4
-- | -- | -- | --
  |   |   |  
  |   |   |  
  |   |   |  
  |   |   |

df8oe:
Cannot confirm here. All is working as expected....

ly2bok:
https://youtu.be/U9EVvJgp-s0
Carefully watch the video. Settings do not change. The level of the signal is SSB, then I switch AM. I return back to the SSB where the signal is much weaker. I turn off the TRx. Again I turn it on. The signal of the SSB is again in a high level ... Agree, it should not be so. All the standard setting. On the other two TRx, that's allso. Ver. 0.4. STM with more memory 1 MB.
Justas 73!

DD4WH:
Hi Justas,
I cannot help you if you do not tell us your AGC setting . . .

DD4WH:
. . . and concluding from your video it seems absolutely normal behaviour because you have switched on the Hang function in AGC.
Please have a look here:
https://github.com/df8oe/UHSDR/wiki/Automatic-Gain-Control-(AGC)
there you will find an explanation of the hang function and how to set the parameters.
All the best 73s
Frank

ly2bok:
I translated the letter from LY3BD Aleksandra. He broadly understands this. I do not know how accurately Google translates, but I think it will be clear. I also enclose the original text in Russian.
[...]

PR #1476, opened 2018-03-16
potential fix for AGC issue in AM/SAM #1438

DD4WH:
untested, please test carefully with AM/SAM demodulation and bandwidths below and above 4k8

* AGC buffer size was 96 (I & Q)
* attach buffer size was 48 @ 12ksps sample rate (decimation-by-4)
* when we increase sample rate for filter bandwidths of > 4k8, the attack buffer size increases to 96 and is then equal to AGC buffer size 

--> Thus, I increased buffer size of AGC buffer to 192 (* 2) 
Maybe this helps solve the problem?

### DIFF
--- a/mchf-eclipse/drivers/audio/audio_driver.c
+++ b/mchf-eclipse/drivers/audio/audio_driver.c
@@ -50,7 +50,7 @@ typedef struct
 	//#define MAX_N_TAU           (8)
 	//#define MAX_TAU_ATTACK      (0.01)
 	//#define RB_SIZE       (int) (MAX_SAMPLE_RATE * MAX_N_TAU * MAX_TAU_ATTACK + 1)
-#define AGC_WDSP_RB_SIZE 96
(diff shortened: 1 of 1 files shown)


WB8HSL:
Installed 2.9.15 this afternoon.

Strong AM signals still de-sense the receiver.  If I cycle through the BW's, the sensitivity is restored.  

Testing is now underway on SSB signals.

I am not sure if this is the FW version with the AGC update test.  Can someone please verify?

Thank you,

Scott 
WB8HSL

DD4WH:
Hi Scott,
you tested the older version.
Have a try with the newest version available since two hours now.
It seems the bug is gone now.
73 Frank DD4WH

WB8HSL:
I have been testing 2.9.16 over the last 24 hours in both AM and side band modes. 

I think the bug is fixed.

 Thank you very much, to all.  

(If I knew a paypal address I would wire some $ for a good German ale, as promised.)

Scott

PR #1485, opened 2018-04-01
Minor things

db4ple:
Split python code into API module and APP module the command line tool is now named uhsdr_tool
Added mcHF 0.7 bootloader install


df8oe:
What's the problem - why is it shown to me as "conflicting"?

db4ple:
There is a "conflict" in uhsdr_tool.py, I don't understand really why this is a conflict (it is an insertion of a comment). But in any case, you can merge it and keep the comment please. Unfortunately, I did not notice this problem before I committed the second part so I could not resolve it on my own easily.

df8oe:
Yes. It is quite easy to solve - I thought you have an idea how it was born...

df8oe:
Now I have seen! The PR was against branch auto-testing which does not exist here...

PR #1486, opened 2018-04-02
Several changes around configuration

db4ple:
Digital Modes are now per VFO & Band. The VFO now remembers which digital mode you had on a given band. Different VFO same band or different band can have different mode enabled. Switch will go to the digital mode last used on this VFO and band. E.g. VFO A on 80m is RTTY, VFO B is PSK, on 40m VFO A is SSB and VFO B is FreeDV. 

We store this permanentely with the other VFO data. 

Internal change: The handling of enabled bands has changed to use a dynamic approach, which allows to enabled or disabled bands based on the hardware capabilites connected to the UI. Right now this doesn't make difference for users, but simplified the code and is for future OVI40 RF HW a must.


df8oe:
There is a problem at small build (freeDV not included - variable undeclared.

db4ple:
Problem fixed.

PR #1510, opened 2018-04-24
VKeypad. Change colour of pressed button text, some other minor changes.

sp9bsl:


DG9BFC:
aehem ... why not ALL buttons are changed in colour?!?
or is it just for testing things out?!?

sp9bsl:
Hi Sigi,
works as expected (at least for me).  I changed the green text on pressed button to white to make it better visible. Did you expected some other changes?

DG9BFC:
check dsp buttons ... why is lower left in yellow??

sp9bsl:
https://github.com/df8oe/UHSDR/wiki/Operating-Manual#dsp-virtual-keypad
Yellow means WARNING, it is intented.

DG9BFC:
ah ok ... that is if you have spi display (that slows dsp down) ... with parallel display it works fine and so no warning needed .... what if you only change that colour if a spi display is used?!? ... with a parallel display the dsp works with nr + notch with no problems ... and in the info screen you can see what display is used ... so warning colour could be made active only if a spi display is  used (ok ok it is just cosmetics)

sp9bsl:
Yes, you're right. Will add some code to check this.

PR #1518, opened 2018-05-09
Updated Codec2 to SVN r3590

db4ple:
May not build, travis had problems (no output shown).


df8oe:
... not critical. Just one warning:

#####
drivers/freedv/ofdm.c:91:18: warning: 'uw_ind' defined but not used [-Wunused-const-variable=]
 static const int uw_ind[]         = {19,20,37,38,55,56,73,74,91,92};  /* index in modem frame of UW bits */
#####

db4ple:
Hi,
the warning is normal, Travis just did do a working build, so I did not knew if all builds go through without problems, but apparently they do.
That warning is probably some work in progress of the Codec2 development

PR #1519, opened 2018-05-10
Oscillator Modifications

db4ple:
Please test before releasing, I was not able to do so for mcHF.

- Si5351a now uses CLK2 for 4x clock , CLK0+CLK1 for 90 degreed phased output @ 1x clock
- Si5351a code now toogles gpio (high for 4x none-phased output, low for 1x phased output
- Oscillator Interface now expects 1x frequency value as input and internally scales by 4
  if is driving a 4x johnson counter based clock phasing circuit.


df8oe:
Many thanks - it works :)

PR #1528, opened 2018-06-03
Encoder boundary crossing fixup. Attempt 1.

sp9bsl:
Tested on mcHF not on OVI40. Added filtering for timer inputs used for encoders. Works perfect with my 256/rotation encoder.
I'm very curious the effect of this fixup. This is strange but there was NO detection against overflow/underflow of timer in encoder mode. Also corrected strange ARR value for all timers used for encoders.
All this is strange to me that no one noticed it before... Anyway, hope it will work for everyone :)
Changes not added to H7 sources/includes.

### DIFF
--- a/mchf-eclipse/basesw/mcHF/Inc/tim.h
+++ b/mchf-eclipse/basesw/mcHF/Inc/tim.h
@@ -62,7 +62,9 @@ extern TIM_HandleTypeDef htim5;
 extern TIM_HandleTypeDef htim8;
 
 /* USER CODE BEGIN Private defines */
-
+#define ENC_filter 15									//encoder input sampling filter (amount of same input samples to be threated as stable state)
+#define ENC_ClockDivision TIM_CLOCKDIVISION_DIV4		//clock divider for encoder filter
+#define ENCODER_RANGE	0xFFF							// Maximum pot value
 /* USER CODE END Private defines */
 
 extern void Error_Handler(void);
--- a/mchf-eclipse/basesw/mcHF/Src/tim.c
+++ b/mchf-eclipse/basesw/mcHF/Src/tim.c
@@ -65,17 +65,17 @@ void MX_TIM3_Init(void)
   htim3.Instance = TIM3;
   htim3.Init.Prescaler = 0;
   htim3.Init.CounterMode = TIM_COUNTERMODE_UP;
-  htim3.Init.Period = 4103;
-  htim3.Init.ClockDivision = TIM_CLOCKDIVISION_DIV1;
+  htim3.Init.Period = ENCODER_RANGE;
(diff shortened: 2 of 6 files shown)


df8oe:
Congrats. That seems to be the goal. I am not able to reproduce the issue any longer...

db4ple:
Great. I noticed the missing handling of "overflow" when I looked at it flowing our discussion in the forum. In fact, there was some code which I suspect should handle this, but I never understood how it was supposed to work. That was the only candidate issue inside the software I say in the software but I was not thinking this would cause the issue so often (as it happens only every 4103 steps or 160 full turns). Probably because we are turning the menu encoders up and down we were crossing the boundary quite frequenqtly, now that I think of it (we never got far away from the 0 value for the encoder 2 probably). 

If this fix works (which seems it does), we need to update the CubeMX configurations to generate the correct configurations (which we need to do if we want/need to get updated HAL libraries in the future).  

Otherwise these configurations become useless. I will do this. This will "replace" the changes in tim.c since this is a generated  file, which we  should not touch if possible.


73
Danilo

sp9bsl:
Hi Danilo, 
the 4103 value is very strange. I have no idea why this number was used (171*24-1). And in somewhere else the 0xfff... Lost of code consistency.
Of course the CubeMX file must be updated like you said. The major thing which could be tested now is the highest possible filter setup  for encoder pins on timers inputs i made. 15 samples and clock divided by 4 for filter stage. My test showed that it works for my encoder 256pulses per rotation and any other low speed i have (24 pulses per rotation). There is part of code with "detent" detection which is in my opinion obsolete but left it, because I do not understand how it should work and it doesn't have negative influence (at least did'n notice that).
Anyway, this is most used part of code and we'll have feedback very soon if something is worse.

73 Slawek

db4ple:
4103 is (0xfff == 4095 + 8).  As I said, we inherited some strange code here, I believe the 8 steps above 4096 were somehow intended to be used for some clever overflow management. When I refactored the code some time ago, I tried to leave the original algorithm intact but this is hard if you don't understand the algorithm entirely. May be I broke it, may it was broken from the beginning. I wouldn't care too much.  Proper overflow handling is a must, yours does the job (hopefully) and then we are ready to go ahead and forget about this issue.

db4ple:
BTW, if the feedback from some more testers is okay, we should remove the old code ASAP. 
Dedetent: I gave that some extra thought already. It simply "eats" small turns (1 to 3 steps).   Some encoders generate 4 steps between each detent. So this code ignores the jumping if you stop accidentically outside of a detent position and then later the encoder "jumps" into place. 
I am absolutely not sure if this code does anything good in the real world. But it does not hurt too much either.

sp9bsl:
Ok, we'll see :)

PR #1534, opened 2018-06-21
Fixed USB Audio device desciptor.

little-tesla:
With this change USB audio works as well on Ubuntu based distros out of the box. bDeviceClass: Most class specifications choose to identify itself at the interface level. This allows for the one device to support multiple classes. bNrChannels: Channel conflict (was 1) and wChannelConfig was set as mono, which won't work as intended. Type III Format: is wrong, it doesn't support PCM at all....so it has to be Type I Format
/HB9GND

### DIFF
--- a/mchf-eclipse/drivers/usb/app/usbd_desc.c
+++ b/mchf-eclipse/drivers/usb/app/usbd_desc.c
@@ -142,8 +142,12 @@ __ALIGN_BEGIN uint8_t USBD_FS_DeviceDesc[USB_LEN_DEV_DESC] __ALIGN_END =
     0x00,                       /* bcdUSB */
 #endif
     0x02,
-    0xEF,                       /*bDeviceClass*/
-    0x02,                       /*bDeviceSubClass*/
+	/* Typically only the bDeviceClass is set at the device level.
+	 * Most class specifications choose to identify itself at the interface level
+	 * and as a result set the bDeviceClass as 0x00.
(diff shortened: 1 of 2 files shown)


db4ple:
Well spotted! Kind of interesting that the other OS are not trouble by the slightly wrong descriptors.

Did you try this on a non-Ubuntu machine as well (Windows and/or MacOS)?

73
Danilo

df8oe:
I cannot test on Windows/Mac but on different Debian based Distributions (not Ubuntu). It is working out-of-the-box as before with one slight difference in software settings: now you have to choose same device for RX and TX - no longer different devices. Kernel 4.16.x.

little-tesla:
Good morning,

sorry I forgot to mention on what platforms the fix was tested.
- Linux Mint (Most recent version, with everything up-to-date) works fine (fldigi, wsjt-x and native audio)
- Windows 7 Pro 64 (System up to date [else device will not be recognized propperly because of a bug in the composite device driver]; STM CDC driver installed and inf file modified as described on the wiki) 
Driver installs fine and I see audio input. I did not test it with any ham-radio software.

img_20180622_072406

I don't have access to an Mac device, so couldn't test it.

Thx and best regards,
Dimce

df8oe:
We will see but I do not think there will be complains. Some OS are more error redundant (Debian) and some not (Win7/64pro, Ubuntu) and from theory I think this fix will work on problematic OS.

73
Andreas

Issue #1546, opened 2018-07-21
[solved] Spurious receive firmware solution? YES! Start with clean (default) settings!


WD8BXS:
Through much experimentation, I have just figured out that the spurious receive (12khz) stated with firmware 2.9.16

I loaded 2.9.15 and the receive is fine.

There is a firware issue somewhere.

73,
 Chuck

df8oe:
I do not think so. I am runing all firmwares up to the newest on many different radios (mcHF and OVI40) and I cannot confirm the issue. I think your configuration is scrambled...

You can test it (and help us to make firmware more tough against malsettings):
First install firmware 2.9.43
Second please make a backup of your current settings via USB to PC. If it IS a configuration problem you can send us this file for bug hunting.
Third start mcHF with default settings and test again if the issue persists. I think not...

73
Andreas

db4ple:
Hi Chuck,
thanks for the testing!

Like Andreas I still not understand why only you and only now have this issue. What is the differentiator between you and the rest of the users so that you have this issue and can reproduce it and others don't. 

We really should look at the  configuration as part of the problem.  For that we need the configuration data (see above and my forum post).

Now looking at the change between 2.9.15 and 2.9.16: It is not much, actually we only increased the size of a datastructure (technically two of them, but nevermind) and some code which initializes this data with zeros when the AGC startup code runs (e.g. when changing modes). 

I cannot see how this code directly could cause the mentioned issue. But still, it could have unwanted side effects especially since we increase the use of RAM (we add 96*12 ~ 1k RAM). Normally this if this is the root cause, other users should also see the issue since this is code used by all mcHF users who use the AGC. But now let us verify that this is not an issue. Please use a "mchf_small" build (it uses much less RAM in addition to using less flash space) and see if the issue is still there.

Please also report RAM size of the device(s) showing the issue.

WD8BXS:
Unclear of how to do the backup to pc. Is this a new feature?

Sent from Yahoo Mail on Android 
 
  On Sun, Jul 22, 2018 at 2:49 AM, DF8OE<notifications@github.com> wrote:   
I do not think so. I am runing all firmwares up to the newest on many different radios (mcHF and OVI40) and I cannot confirm the issue. I think your configuration is scrambled...

You can test it (and help us to make firmware more tough against malsettings):
First install firmware 2.9.43
Second please make a backup of your current settings via USB to PC. If it IS a configuration problem you can send us this file for bug hunting.
[...]

PR #1569, opened 2018-08-27
Set all TX Power defaults to 0 which is a safe default.

db4ple:
With the switchable scaling of tx power config values, "real" defaults make no sense in the current implementation since the config mechanism currently uses fixed default constants (which are not changed when you change the scaling back and forth).
There are ways to solve this, but using 0 is safe.


db4ple:
This is a result of the discussion about "clean" config reset in #1546 and also here https://www.amateurfunk-sulingen.de/forum/index.php?board=15;action=display;threadid=1161
Thanks Chuck!

Issue #1594, opened 2018-10-28
Issues with Hamlib and WSJT-X


phaethon:
I have been debugging issues with wsjt-x, which sometimes disconnects (requiring reconnection and re-enablement of TX). Also, there is an issue with frequency, which sometimes jumps. 

I have submitted some patches to hamlib (which are incorporated in last version), but there are some issues, which seem to be at least partly on UHSDR side:
- hamlib tries to read eeprom from address 0x0065 and gets no response from UHSDR
- when in tx mode hamlib uses tx status (command 0xf7) to determine if rig is in split mode. UHSDR in tx mode returns 0x00 always, disregarding split status
- after switching from PTT ON to OFF, there is a delay on UHSDR side. hamlib expects immediate switching, which sometimes leads to a situation that hamlib thinks it is disconnected

I have patched wsjt-x/hamlib to work. Patches are easy (when you find out what to do). More complex is the question what to do on UHSDR and what on hamlib side. Which requires discussion.

db4ple:
Hi,

EEPROM@0x65: according to http://www.ka7oei.com/ft817_meow.html and after looking into the Hamlib Github the bits of interest to Hamlib are the bits defining the digital mode.  Not sure exactly how we do this but I think we can at least return USER-L or USER-U. Not sure how relevant it is to report other modes. FT817 does not differentiate between RTTY-L and RTTY-U as we do in the UHSDR. So it may be better to report the sideband instead of not doing. Someone has to figure out what is better. Anyway, I'll fix that, no problem.

F7 -> Easy to fix.

PTT ON to OFF -> not much we can do on that on our side. It takes the time it takes. I don't think we should report the false status. Looking at  the code, I don't see how long they are waiting for us to respond with the expected status (I can see that the request sent multiple times if we not return the expected outcome).
I think, Hamlib should wait a little longer to fix this issue. Not much harm on the Hamlib side. But of course, it would be a fix for a device which just pretends to be a FT817 so these guys may reject it and we could not blame them for that. But worth a try.

phaethon:
It is important to return at least something to 0x65 read, otherwise hamlib is waiting for response, which causes all sorts of wrong effects. Actually, it might be good idea to respond by default with 0x00 to any eeprom read.

[...]

PR #1596, opened 2018-10-28
Swap PTT state logic to match Hamlib

miek:
Details in #1594

### DIFF
--- a/mchf-eclipse/drivers/cat/cat_driver.c
+++ b/mchf-eclipse/drivers/cat/cat_driver.c
@@ -1285,14 +1285,14 @@ static void CatDriver_HandleCommands()
         case FT817_PTT_STATE:
         {
             uint8_t tx_state = limit_4bits(roundf(swrm.fwd_pwr));
-            tx_state |= is_splitmode() ? 0x00: 0x20;
-            tx_state |= ts.txrx_mode == TRX_MODE_TX ? 0x80 : 0x00;
+            tx_state |= is_splitmode() ? 0x20 : 0x00;
+            tx_state |= ts.txrx_mode == TRX_MODE_TX ? 0x00 : 0x80;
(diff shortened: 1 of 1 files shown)


mchf-eclipse/drivers/cat/cat_driver.c
```diff
@@ -1286,10 +1286,10 @@ static void CatDriver_HandleCommands()
...
             tx_state |= is_splitmode() ? 0x00: 0x20;
-            tx_state |= ts.txrx_mode == TRX_MODE_TX ? 0x80 : 0x00;
+            tx_state |= ts.txrx_mode == TRX_MODE_TX ? 0x00 : 0x80;
             tx_state |= swrm.vswr_dampened > 3.0 ? 0x40 : 0x00;
 
-            resp[0]= ts.txrx_mode == TRX_MODE_TX ? tx_state : 0x00;
+            resp[0]= ts.txrx_mode == TRX_MODE_TX ? tx_state : 0x80;
             if(RadioManagement_IsTxDisabled())
             {
                 resp[0] = 0x00;
```
db4ple: Please let the code return 0xff here as it was in the original code.

miek:
I've re-pushed the commit to return 0xFF in receive as requested, I've also changed it to return 0xFF when TX is disabled as I missed that before.

While testing I noticed that WSJT-X thought I was in split mode, and I found that the logic was swapped there too so I've changed that also. https://github.com/Hamlib/Hamlib/blob/master/yaesu/ft817.c#L567

db4ple:
Thanks!  Just saw this myself, when checking the KA7OEI against hamlib: the whole bits on the KA7OEI page are the wrong way it seems. Of course Hamlib is right and our old code was also somewhat right (if not in split mode).

PR #1598, opened 2018-10-31
Towards flexible translate frequency

db4ple:
Removed all direct references to the non-zero frequency translation mode,
replace with use of actual translation frequency.
This change alone does not permit free selection of translation frequency in the full
possible range (+/-24khz) since we have in some places code assume the translation frequency
being a in a power of 2 relationship with the sample frequency of 48000 khz.
But it makes it much easier. I think due to missing performance we should
not enabled free choice of translation frequency on the STM32F4 (here 12khz is
best for performance) but the F7 and H7 should have enough horsepower to
use arbitrary translation frequency in the baseband.

This change should not cause a noticable difference. If it does,
it is unintended and can be fixed.

### DIFF
--- a/mchf-eclipse/drivers/audio/audio_driver.c
+++ b/mchf-eclipse/drivers/audio/audio_driver.c
@@ -1620,10 +1620,10 @@ static void AudioDriver_NoiseBlanker(AudioSample_t * const src, int16_t blockSiz
 //*----------------------------------------------------------------------------
 void AudioDriver_FreqConversion(float32_t* i_buffer, float32_t* q_buffer, int16_t blockSize, int16_t dir)
 {
-    static bool recalculate_Osc = false;
     // keeps the generated data for frequency conversion
     static float32_t                   Osc_I_buffer[IQ_BLOCK_SIZE];
     static float32_t                   Osc_Q_buffer[IQ_BLOCK_SIZE];
+    static int32_t conversion_freq = 0;
 
     assert(blockSize <= IQ_BLOCK_SIZE);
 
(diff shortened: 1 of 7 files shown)


DD4WH:
Hi Danilo,
it is not clear to me what the intention of these changes is. But maybe it is just the wording that irritates me:
* the IF should in most cases be samplerate/4 in order to use cos (1,0,-1,0) and sin(0,1,0,-1) to "multiply"; at the moment there are still hard-coded values of 1500 or 6000 which probably refer to a fixed sample rate of 48kHz (as far as I understand)
* if we want more flexible tuning by means of a software oscillator (in order to have a fixed spectrum and tune around in the fixed spectrum), I would call it "software tuning" or similar, but there are some serious issues with that (no AM demodulation in the centre, when IF = 0 etc.)
So, could you explain a little more what the intention of these changes is?
73 Frank

db4ple:
Hi Frank,
you are right, samplerate/4 is preferred due its highly-optimized calculation. And of course close to IF = 0 is also a problem.

The changes should help us with
* support for multiple iq sample rates 
* support for "free selection" of the shift frequency in a given useful (!) range. This makes more sense with higher sample rates but still should work for our +/-24 khz ranges. In magnify 2x it would allow you to basically go almost from left to right of the spectrum without shifting the spectrum view. For 1x it makes not too much sense I would think since we have this "middle" area around IF = 0 as you mentioned.
* it also helps us if we want to implement two independent receivers in the same spectrum (which, to be honest, would be quite a task, and these changes are only a very tiny part of it).
* it makes in some places the code easier to read (at least for me).   
* it should not do any harm (well, I already found out, I did a small mistake, which forces permanent recalc of the sin/cos table but this is fixed easily)

DD4WH:
Hi Danilo,
thanks for your clarification!
73 Frank
P.S.: two ears --> two simultaneous receivers, why not ;-)

df8oe:
Quisk does have the option of two receivers from the beginning and I like it...

db4ple:
Two receivers is quite a long way to go since large parts of our receive chain are coded under the assumption of having a single receive path. But it is doable and mostly carefull review and refactoring work, not so much basic changes to the algorithms.

sp9bsl:
Hi guys,
I tried to use two receivers on Zeus Radio and my HiQSDR, but it blows away the brain with two ears on headphones :)

db4ple:
Then you need more ears ;-) 1 receiver -> two ears, 2 receivers -> 4 ears, that should work. Especially if you also use two brains. I am at the moment not sure what you could use 2 receivers for except for digital mode use cases where one receiver is tuned to some digital transmission like RTTY and the other can be used to take part to some kind of conversions on an audio mode. But it would be cool, wouldn't it....
We have to find something to eat up all the horsepower the STM32H7 gives us.

df8oe:
My first idea to use horse power and RAM in F7 and H7 was to tun OVI40 into a whisper beacon. ARM processor which is present in Red Pitaya is much more powerful than H7 - but processor load by decoding 5 bands simultaneously is not higher than 5%. TX should not be problematic at all.

db4ple:
BTW, I tested to use arbitrary translation offsets (I used 9000 und -20100) and it seems to work quite well.
The spectrum showed (in 1x mode) the translation at the correct point and the FT8 receive and spectrum was working flawlessly if I switched the translation frequencies between 12000, 20100 and -9000.

PR #1618, opened 2018-11-21
Fix makefile

m-chichikalov:
I failed to build under the windows. 

GNU Make 4.2.1
Built for x86_64-w64-mingw32
Windows 7 x64
GNU Tools for Arm Embedded Processors 7-2017-q4-major - 7.2.1 20170904 
MINGW64_NT-6.1

It's very known issue with windows...
There are the next fixes:
1. all objects passing to linker as one file this list of them (windows has limitation on number of parameters and length of parameter for creating process, so...)
2. As long as I know windows cannot run .py files as executable, so needed mention explicitly. e.x. "python somefile.py"
3. There is no default values for FixPath in Makefile. So cleaning faild on my system.

### DIFF
--- a/mchf-eclipse/Makefile
+++ b/mchf-eclipse/Makefile
@@ -316,13 +316,14 @@ $(FIRMWARE).elf:  $(HAL_OBJS) $(DSPLIB_OBJS) $(OBJS)
 	$(ECHO) "  [LD] $@"
 	@$(file > $(ROOTLOC)/firmware_obj_list.lst,$^)
 	@$(CXX) $(LDFLAGS)  -Xlinker --gc-sections -Llibs -Wl,-Map,$(FIRMWARE).map -o$@ @$(ROOTLOC)/firmware_obj_list.lst $(LIBS)
+	$(RM) $(call FixPath,firmware_obj_list.lst)
 
 # compilation
 $(BOOTLOADER).elf:  $(BL_HAL_OBJS) $(BL_OBJS)
 	$(ECHO) "  [LD] $@"
 	@$(file > $(ROOTLOC)/bootloader_obj_list.lst,$^)
 	$(CXX) $(LDFLAGS) -Xlinker --gc-sections -Llibs -Wl,-Map,$(BOOTLOADER).map -o$@ @$(ROOTLOC)/bootloader_obj_list.lst $(LIBS)
-
+	$(RM) $(call FixPath,bootloader_obj_list.lst)
 
 $(TRX_ID).handbook:
 	@$(ROOTLOC)/support/ui/menu/mk-menu-handbook auto
--- a/mchf-eclipse/Makefile
+++ b/mchf-eclipse/Makefile
@@ -196,7 +196,7 @@ OS = @${PREFIX}/bin/arm-none-eabi-size
 HEX2DFU = $(ROOTLOC)/support/hex2dfu/hex2dfu.py
 
 ifdef SystemRoot  # WINxx
-	HEX2DFU = python $(ROOTLOC)/support/hex2dfu/hex2dfu.py
+    HEX2DFU = python $(ROOTLOC)/support/hex2dfu/hex2dfu.py
     CC = arm-none-eabi-gcc
     CXX = arm-none-eabi-g++
     OC = arm-none-eabi-objcopy
@@ -204,7 +204,7 @@ ifdef SystemRoot  # WINxx
 endif
 
 ifdef MSYSTEM
-	HEX2DFU = python $(ROOTLOC)/support/hex2dfu/hex2dfu.py
+    HEX2DFU = python $(ROOTLOC)/support/hex2dfu/hex2dfu.py
     CC = arm-none-eabi-gcc
     CXX = arm-none-eabi-g++
     OC = arm-none-eabi-objcopy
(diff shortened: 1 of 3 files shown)


sp9bsl:
Hi,
hmm I did not find any problems with build under the windows as yet...

m-chichikalov:
Ups, looks like it failed on Travis checks... I will check later.

> Hi,
> hmm I did not find any problems with build under the windows as yet...

Hi, could you describe your toolchaine setup, please?

sp9bsl:
[In our Wiki](https://github.com/df8oe/UHSDR/wiki/Setting-up-Firmware-Development-Software#gnu-mcu-eclipse-formerly-gnu-arm-eclipse) you can find the description how to install the GNU MCU ECLIPSE which works just fine.

db4ple:
I run my "Windows" makefile build under "WSL" (the Windows Subsystem for Linux), that is the most painfree thing, since it runs directly on the same directory my Windows Eclipse runs, so I have best of both worlds. Much easier to setup than MinGW it turned out.

Anyway, fixing it for MinGW would be nice, too.

df8oe:
Yes there is a remaining problem...

Regarding Linux CLI integrated in Windows: I have seen many reports for that and all say: "It is nice but 1) There are many command that are not implemented and 2) The ones that are implemented are sometimes incompatible with the "original". Means: you can use them, but syntax will not work for Linux machines. I do not hope that is the cause of the problem here... We'll see!

Very sad that Windows CLI for Linux commands has same restriction to filenames/paths. That is one of the incompatibilities...

db4ple:
@df8oe: WSL is a full Linux (running on top of a Linux kernel which runs on top of the MS Windows kernel ! WSL is (in my case)  an Ubuntu install with a terminal emulator (you can have other distros as well). That is absolutely what I want for compilation.  The command line has no Windows restrictions, it runs in Linux and unless you look at the devices, there is no difference.

I guess what you are refering to is MinGW and/or Cygwin, which emulate kind of Unix-ish/posix interface and to a certain extent have the limitations of native Windows as they translate the call to the Windows API directly.  MinGW uses the normal Windows CMD prompt  and Windows API for process execution and has for that reason the same command line length restriction AFAIR. 
So these are close but not perfect.  

But nothing speaks against the passing of the whole object list as a single parameter. That would be nice, and would not create issues under Linux as well if done right. Thumbs up for this 👍 

@m-chichikalov: You have to generate the list in the build directory or below, so that individual builds can have different list, see .travis.yml for how we build multiple variants in separate directories. Should not be difficult.

m-chichikalov:
@db4ple, sure I will check and rework. Thank you.

m-chichikalov:
> WSL is a full Linux (running on top of a Linux kernel which runs on top of the MS Windows kernel ! WSL is (in my case) an Ubuntu install with a terminal emulator (you can have other distros as well). That is absolutely what I want for compilation. The command line has no Windows restrictions, it runs in Linux and unless you look at the devices, there is no difference.

Nice news, because the compiling under CYGWIN is much slower... I should test it. Thank you. I didn't know about WSL.

mchf-eclipse/Makefile
```diff
@@ -308,13 +314,16 @@ $(BOOTLOADER):  $(BOOTLOADER).bin $(BOOTLOADER).dfu
...
+	@$(CXX) $(LDFLAGS)  -Xlinker --gc-sections -Llibs -Wl,-Map,$(FIRMWARE).map -o$@ @$(ROOTLOC)/firmware_obj_list.lst $(LIBS)
+	$(RM) $(call FixPath,firmware_obj_list.lst)
 
 # compilation
 $(BOOTLOADER).elf:  $(BL_HAL_OBJS) $(BL_OBJS)
 	$(ECHO) "  [LD] $@"
-	$(CXX) $(LDFLAGS) -Xlinker --gc-sections -Llibs -Wl,-Map,$(BOOTLOADER).map -o$@ $^ $(LIBS)
-
+	@$(file > $(ROOTLOC)/bootloader_obj_list.lst,$^)
+	$(CXX) $(LDFLAGS) -Xlinker --gc-sections -Llibs -Wl,-Map,$(BOOTLOADER).map -o$@ @$(ROOTLOC)/bootloader_obj_list.lst $(LIBS)
```
db4ple: Hi, I tried this construct locally and it worked for me. I have no idea yet why it does not work in the travis environment. Nevertheless we should make it work there. In any case, leave out the $(ROOTLOC) and the path issue I was mentioning should be resolved. If this also fixes the travis issue, I have no idea

db4ple:
Hi,
just found this, maybe the installed travis make is too old:
https://stackoverflow.com/questions/23548651/using-makes-file-function

db4ple:
Yes, this is it:
https://packages.ubuntu.com/trusty/make is 3.81
$(file ) is only in 4.0

db4ple:
Change 
dist: trusty to dist:xenial  in .travis.yml should do the trick

m-chichikalov:
> Change
> dist: trusty to dist:xenial in .travis.yml should do the trick

Thank you, will try.

db4ple:
However on Linux the `@echo '$^' >file.lst` approach would work always even with older makes.
Maybe we should do two approaches to not risk any Linux user of not so "advanced" distros to become unhappy.

m-chichikalov:
> However on Linux the 1@echo '$^' >file.lst1 approach would work always even with older makes.
> Maybe we should do two approaches to not risk any Linux user of not so "advanced" distros to stay happy.

Good idea.

m-chichikalov:
> However on Linux the 1@echo '$^' >file.lst1 approach would work always even with older makes.
> Maybe we should do two approaches to not risk any Linux user of not so "advanced" distros to stay happy.

Now echo failed because of length of parameters :)

db4ple:
Remember, on Windows you cannot pass the '$^' to any command line tool, it cannot handle this length. That is what this change is all about...

m-chichikalov:
Ok, with another version of make (in another distr.) the Travis can handle this approach. Thank you @db4ple

df8oe:
make package in Debian Jessie (which is the oldest we want to support) is only 4.0 - even with added backports repositories. So make in version 4.0 should be supported. Good approach!

Issue #1625, opened 2018-11-27
Discussion:  1.Display module mess, 2. Added RTOS.


m-chichikalov:
I would like to discuss couple topics, not sure that this is the best place for it, if not let me know where is it...
Bear with me guys, please.

### "Displays mess" 
Do we really need to check (recognize) the display on the run time?  We would end up that even 2mb is not enough. Maybe better to compile several different binaries for each supported display? Or I do not understand something? 
As long I want to add support for another display, I see that there is such big mess in a display module.
Seems like some steps were done to be abstract but not 100%

### "All done in INT context."
Another point what I do not like that everything, almost everything, is done in interrupts, so it's a trap guys... There is no way to easily maintenance a system like this.  What do you thinking about some RTOS? I'm sure .... someone already did it and got that there is no enogh power resources of CMU or memmory? I do not have mcHF HW but I saw videos and number of MCU usage on the screen was around 34% this is true?  

I know, when in Rome, do as the Romans do, however...
Does someone have thoughts about it? Share please, let's discuss

PS. Maybe do you have another place for topics like this?

sp9bsl:
Hi Max,
I will shortly answer you with the "display mess" because I'm partially responsible for this. The unified driver we have today has one advantage: one executable for all machines with the same core, regardless of used display. In our opinion this way has an advantage becuse creates less mess for the users. Of course the code might be worse or better written, we continously work at it, you are welcome if want to help. Currently there is unified driver for 320x240 and 480x320 and almost ready for 800x480. The last one will be available only for F7/H7 cores.

I highly recommend you to have mcHF or better OVI40 UI hardware before introducing any changes to code. 
The best place to discuss new/improved ideas is[ our forum](https://www.amateurfunk-sulingen.de/forum/). Please register there and we can discuss, this tab is mainly for resolving the issues.

73 Slawek

m-chichikalov:
> I highly recommend you to have mcHF or better OVI40 UI hardware before introducing any changes to code.

Sure, that's why I did not touch anything yet )))

> The best place to discuss new/improved ideas is[ our forum](https://www.amateurfunk-sulingen.de/forum/). Please register there and we can discuss, this tab is mainly for resolving the issues.

I will, thanks for the link.

> 73 Slawek

73!

db4ple:
Hi,
a couple of things:
[...]

PR #1626, opened 2018-11-28
Fixed display driver build an issue reported in #1625

db4ple:
Found some not properly initialized loops in ui_spectrum.c
Turned on some more compiler warnings and worked fix some in this commit.
Many more warning fixes to come in future pull requests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,7 @@ script:
     - cd ../build-bl-f7
     - make -f ../Makefile OPT_GCC_ARM=$GCC_DIR ROOTLOC=".." BUILDFOR="F7" TRX_ID="ovi40" TRX_NAME="OVI40" bootloader
     - cd ../build-fw-f4-ili9486-480
-    - make -f ../Makefile OPT_GCC_ARM=$GCC_DIR ROOTLOC=".." LCD_TYPE=1 all
+    - make -f ../Makefile OPT_GCC_ARM=$GCC_DIR ROOTLOC=".." CONFIGFLAGS="-DEXTERNAL_USE_GFX_CONFIG -DUSE_DISP_480_320 -DUSE_GFX_ILI9486" all 
     - cd ../build-fw-f4-small
     - make -f ../Makefile OPT_GCC_ARM=$GCC_DIR CONFIGFLAGS="-DIS_SMALL_BUILD" ROOTLOC=".." all
     - cd ..
--- a/mchf-eclipse/hardware/uhsdr_board.h
+++ b/mchf-eclipse/hardware/uhsdr_board.h
@@ -73,10 +73,16 @@
 // this is the standard NR now (Febr 2018)
 #define USE_ALTERNATE_NR
 
-
+// you may optionally define the list of supported GFX drivers externally
+// if you just define EXTERNAL_USE_GFX_CONFIG and no USE_GFX_...
+// you will get a headless system using a dummy display driver
+#ifndef EXTERNAL_USE_GFX_CONFIG
   #define USE_GFX_ILI932x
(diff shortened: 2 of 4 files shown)


mchf-eclipse/hardware/uhsdr_board.h
```diff
@@ -73,10 +73,16 @@
 // this is the standard NR now (Febr 2018)
 #define USE_ALTERNATE_NR
 
-
+// you may optionally define the list of supported GFX drivers externally
+// if you just define EXTERNAL_USE_GFX_CONFIG and no USE_GFX_...
+// you will get a headless system using a dummy display driver
+#ifndef EXTERNAL_USE_GFX_CONFIG
```
sp9bsl, line 79: Hi Danilo,
what is purpose of the headless system?

.travis.yml
```diff
@@ -44,7 +44,7 @@ script:
     - cd ../build-bl-f7
     - make -f ../Makefile OPT_GCC_ARM=$GCC_DIR ROOTLOC=".." BUILDFOR="F7" TRX_ID="ovi40" TRX_NAME="OVI40" bootloader
     - cd ../build-fw-f4-ili9486-480
-    - make -f ../Makefile OPT_GCC_ARM=$GCC_DIR ROOTLOC=".." LCD_TYPE=1 all
+    - make -f ../Makefile OPT_GCC_ARM=$GCC_DIR ROOTLOC=".." CONFIGFLAGS="-DEXTERNAL_USE_GFX_CONFIG -DUSE_DISP_480_320 -DUSE_GFX_ILI9486" all
```
sp9bsl, line 47: Danilo, with this you disabled the travis check for 320x240 driver and for my new 800x480

sp9bsl: Sorry for late comments, forgot to press finish review :)

db4ple: Hi Slawek,
how can this change disable your check? There is no active reference in the current code which relates to LCD_TYPE since you deleted the respective statements in uhsdr_board.h.

If you are using it for some purpose,  I was not aware of this. 
Easiest and safest solution: simply add the respective statement in your branch after rebasing. This will keep these changes in even in future rebase action.

I would still want to have a build with not all display drivers enabled as part of the travis. 

BTW, the headless build is just a by-product of our approach, not an explict goal. I just noticed that we have that compiling without a problem. But in fact using such a binary is not much different from one with the wrong display drivers compiled in, you'll get the signal that no display has been detected.

sp9bsl: hmm, not tested so far of course, but defining the EXTERNAL_USE_GFX_CONFIG with only 480x320 driver you disabled the check for other drivers by undefining them in uhsdr_board.h line#79. Why do you need to disable display drivers check for travis?

Edit: ok, I see earlier build for f4 with "all" option, but why is the second build for f4 with 480x320 only?

db4ple: Hi Slawek,
this is only one of multiple builds we run inside the single travis job, have a look at the full .travis.yml 

We cannot test all of the different combinations of individual drivers and resolutions. 
Purpose of this build is to test a minimal configuration using a single driver plus a single resolution. Now that we are discussing, it occurs to me that the USE_DISP_480x320 may make it two resolutions. Anyway, that is the motivation. 
In corner cases we may have to provide for instance stripped down bootloaders with just a single display driver to keep the size it below the limit.

sp9bsl: Yes you're right, I overlooked the previous fw build with the f4, correct me but I think the case of exclusive fw build for f4+480x320 is a relict when we tested the drivers at early stage before unification (that's why i removed the LCDTYPE from defines). But for the whole idea of striping down the bootloaders out of not necessary drivers is understood.

PR #1629, opened 2018-11-29
Now all SPI display drivers and/or parallel display drivers can be disabbled

db4ple:
via #define

#ifdef's were not correctly placed.
No functional change


m-chichikalov:
Now all SPI display drivers and/or parallel display drivers can be disabbled !? Are you sure... I will find something more, just give me time)))

db4ple:
Well, these USE_... which are defined inside the actual implementation files are not ment to be changed as part of the normal configuration (all of these global defines are in uhsdr_board.h) , these are used for development purpose.
But in general you're right, you can run without display drivers enabled. But most developer will quickly notice this. In case not, I will add some warnings being spit out during compile.

PR #1643, opened 2018-12-11
Added VSWR protection of PA

satoryboy:
Activating by parameter "VSWR Protection" in Debug menu. If VSWR > 3, power of tx/tune will be switched down. Interrogation of SWR-meter every 10 ms. TX power limits for VSWR > 3 and VSWR > 5 is adjustable (and just minimum - 0.5W if VSWR > 10; in case controller will not hang due to reflected wave).


satoryboy:
Add clearing of message "VSWR protection worked" on LCD when tuning of TX or TUNE power level.

sp9bsl:
Hi Yuri,
omiting the reason to add this protecion or not, please be aware of the Flash memory size, especially in small build. Using the function  UiDriver_TextMsgPutChar() this way is not code efficient, please use const buffer and a simple loop instead. Also, please consider printing info like the one we have during and a while after eeprom save.

satoryboy:
Thanks for your comments. I'm not good coder.

db4ple:
Hi Yuri,

thanks for submitting another piece of code. I can't say this often
enough, we appreciate what you and others are doing and are open minded.
And we'll help you to become a better coder, if you want.

However, I have to make a few comments about your code (and some general
comments at the end). 

Please do not call directly or indirectly any functions related to the
user interface handling from RadioManagement_...
functions.
Instead create a user function which reads for instance some variables
controlled by the RadioManagement and changes the display accordingly.
There is an theoretical exception to that rule, that is
UiDriver_TextMsgPutChar() since this one is designed to just fill a
buffer and never blocks or causes long delays.
However, the TextMsg is reserved to be used for RTTY, CW etc. where the
received or to be transmitted text is being displayed. It is not meant
to be a "Notification" area. Having said this, using this area in
experimental code is fine, if it is not active by default and is later
rewritten to follow the rule of not using it. 

UiDriver_TextMsgClear() on the other hand, that one calls the display
driver, and we don't do that in RadioManagement functions.
 
These rules are not written anywhere yet, so it is not your fault, but
still worth being obeyed.

Another thing to mention is the use of EEPROM config variables in
experimental code. Unless absolutely necessary, don't store the
experimental configuration permanently in EEPROM.
 
Reason being, an EEPROM parameter number is bascially "burned" for other
use once we used it in a single firmware version for a specific purpose.
If we would reuse it and load a different firmware which uses the same
parameter for a different purpose a lot of strange things can happen.
And if we simple use new numbers all the time, we will "quickly" run out
of numbers and especially would have lot of "dead parameters".

So in general, implement the experimental function with non-stored
variables in the debug menu; see if everyone is happy with it the way it
is; and once finalized and possibly discussed, assign as many EEPROM
parameters as necessary to it for permanent use.

Now some more comments, not as critical as the ones before.

The duplication of the power level choice string selection I don't like
too much. We should refactor the translation of the power level into a
string into its own function and use this everywhere (including the
existing two place where we have a duplication).
Yes, we could have done this before, but it was only two place not four,
where we duplicated code. Not a big deal.

I am also tempted to suggest to move the variables from the ts to the
swrm structure. But this is even less critical.

If you or anyone else now starts wondering, why these rules are not
documented somewhere: Well, writing rules down is a lot of work.
As we look at incoming code, we check if the code is following the basic
rules, and if it does, why should be write down rules normally not broken.
But once more persons starts to work in critical areas and submit code
with issues, we have to act.

However, no one should be force to read through pages of coding rules.
The rules are meant to keep the software quality high, the software code
readable and maintainable, and the supported devices running smoothly.
Maybe we should start discussing how to document these rules now.

BTW, it is absolutely fine to question the rules, and discuss them and
also changing them on mutual agreement and with a reason.

m-chichikalov:
@db4ple  :+1:
Also would be great to start thinking about boundary where we will separate HW dependent code from TRX logic, this is not good to have #if define (F4) every where around the code. 

E.X. Do you remember I'm working on the display, so The logic, should not know even what kind of SPI port is in use by HW, it should call dedicated functions (the same for all HW) in UHSDR\mchf-eclipse\basesw\mcHF\ or UHSDR\mchf-eclipse\basesw\ovi40\ ... or any other....  the same for fsmc or I2C...

What do you think?

db4ple:
Well Maksim,
you touch an very interesting topic.

In general I agree very much, in the particular case, I don't agree. The use of #ifdef to address processor variation is limited to very few places in the mentioned file. Most of the code in said file is written without caring about which processor it runs on. Of course we could factor out these few functions, but what good would it do? 
If the goal is to make the software completely independent from the processor family we use (STM32xxxx) that would be a different story. 

I am absolutely in the boat to minize the use of processor dependent code spread across the whole system. I am not against #ifdef. Just do a text search for STM32H7, STM32F7 and STM32F4, you'll see references to these only in place operating very close to the hardware.

The one where I would agree to have that removed/replaced is in ui_spectrum, here it is more about the available memory than a processor dependency. 

What my goal was when doing the ports to different processors (I did port to F7 and H7) and by this was introducing the #ifdefs was to keep as much code common and shared between processors as possible in the main UHSDR code base (i.e. outside of basesw).  That sometimes defeats the "text book" separation of hardware dependent and hardware independent code. I agree. Still I see more value in following this strategy in the concrete situation we are in. There is simply not enough difference to justify this. I like to have these small differences rather in plain sight than to have multiple similar abstraction functions.

And believe me, if you know how we can get away without unnecessary code duplication AND without #ifdef related to processor choices and same or better performance code in the graphics driver, just do it. As long as it works as good as now, I won't complain. Promise! 

However, I would not want to have it in basesw/xxxx/ because this contains only code as generated from the CubeMX tool. Some of this code has been more or less heavily modified, nevertheless it is reservered for files from this "source". To be honest,  I am not sure how we are going to handle these modifications of generated code anyway, I would like to minimize them further for the same reason as above. But a lot of the setup stuff done there is really very processor specific, not much too reuse possible. 

And if we start from scratch, I would do a lot of things differently, starting with not using C as the main implementation language altogether. But this is a completely different story.

sp9bsl:
Hi Guys,
Yes the complexity of the code increases constantly... With all new hardware approach we have more things to test aganist compatibility. This is because UHSDR supports few platforms, and we are currently working on two new RF setups, one of them causes peripherial reconfiguration and different approach to audio driver, so another ifdefs will appear... My idea is to not touching the basesw but reconfigure the peripherial after CubeMx code does it, I can do it because my hardware allows this (FPGA).
@satoryboy, @m-chichikalov : my comments may look critical but please do not threat it as discouraging, some time ago I also came here and since this moment I've learned a lot thanks to Danilo, Andreas, Frank and others... (and of course still learn).

m-chichikalov:
@sp9bsl, @db4ple thank you for response. 

> Well Maksim,
> you touch an very interesting topic.
> 
> In general I agree very much, in the particular case, I don't agree. The use of #ifdef to address processor variation is limited to very few places in the mentioned file. Most of the code in said file is written without caring about which processor it runs on. Of course we could factor out these few functions, but what good would it do?
> If the goal is to make the software completely independent from the processor family we use (STM32xxxx) that would be a different story.

I'm sure you agree with me that now it's a stretch to call it UniversalHSDR. And by the way, what do you understand under Universal in UHSDR? I came with an idea to port UHSDR under well known sdr but later I realised that there is nothing to call Universal, it's realy close tied to mcHF hardware and hard to port...  this is right, it was born for it.. But we should go toward more and more HW independent code... It will give more easiest way to use UHSDR under more platforms...

> I am absolutely in the boat to minize the use of processor dependent code spread across the whole system. I am not against #ifdef. Just do a text search for STM32H7, STM32F7 and STM32F4, you'll see references to these only in place operating very close to the hardware.
> 
> The one where I would agree to have that removed/replaced is in ui_spectrum, here it is more about the available memory than a processor dependency.
> 
> What my goal was when doing the ports to different processors (I did port to F7 and H7) and by this was introducing the #ifdefs was to keep as much code common and shared between processors as possible in the main UHSDR code base (i.e. outside of basesw). That sometimes defeats the "text book" separation of hardware dependent and hardware independent code. I agree. Still I see more value in following this strategy in the concrete situation we are in. There is simply not enough difference to justify this. I like to have these small differences rather in plain sight than to have multiple similar abstraction functions.
> 
> And believe me, if you know how we can get away without unnecessary code duplication AND without #ifdef related to processor choices and same or better performance code in the graphics driver, just do it. As long as it works as good as now, I won't complain. Promise!

I slowly do, not have to much time... And other problem I do not have all HW, so I'm sure I broke some already, but I will publish binaries for test before make PRs...
And I'm waiting ovi40 kit, it should appear shortly.

> However, I would not want to have it in basesw/xxxx/ because this contains only code as generated from the CubeMX tool. Some of this code has been more or less heavily modified, nevertheless it is reservered for files from this "source". To be honest, I am not sure how we are going to handle these modifications of generated code anyway, I would like to minimize them further for the same reason as above. But a lot of the setup stuff done there is really very processor specific, not much too reuse possible.

I agree. 

> And if we start from scratch, I would do a lot of things differently, starting with not using C as the main implementation language altogether. But this is a completely different story.

I'm working in one project where we using mostly all drivers in C and high level stuff in C++ where are also some problems with that but I agree, in some cases using C++ would be great.


> Hi Guys,
> Yes the complexity of the code increases constantly... With all new hardware approach we have more things to test aganist compatibility. This is because UHSDR supports few platforms, and we are currently working on two new RF setups, one of them causes peripherial reconfiguration and different approach to audio driver, so another ifdefs will appear... My idea is to not touching the basesw but reconfigure the peripherial after CubeMx code does it, I can do it because my hardware allows this (FPGA).

Maybe this is exactly the place where some work should be done to make it more independent?

> @satoryboy, @m-chichikalov : my comments may look critical but please do not threat it as discouraging, some time ago I also came here and since this moment I've learned a lot thanks to Danilo, Andreas, Frank and others... (and of course still learn). 

Don't warry. Thank you...

satoryboy:
Please close (reject) this pool-request. I need timeout to think and experiments with mine rig. And sorry about "quick...". Learning Eclipse.

df8oe:
closed following wish of @satoryboy

PR #1649, opened 2018-12-20
Testing VSWR Protection of PA

satoryboy:
Activated by parameter "VSWR Protect. threshold" in Debug menu (don't store in EPROM, default is "OFF"). Value range is 2...6. If on TX/tune mode VSWR will exceed this value, the bias of PA will down to zero and the LED "TX" will flash quickly until the TX/tune mode is turned off. [Sorry about some garbage commits. Actual is just last commit for actual version 2.9.88]


mchf-eclipse/drivers/ui/ui_driver.c
```diff
@@ -4516,12 +4516,18 @@ static void UiDriver_HandleTXMeters()
         swrm.p_curr   = 0;
         swrm.fwd_calc = 0;
         swrm.rev_calc = 0;
+        swrm.high_vswr_detected = 0;
```
m-chichikalov: if you change it to line below it's more obvious that this variable is type of bool, isn't it?
`
swrm.high_vswr_detected = false;
`

satoryboy:
Thanks!

db4ple:
Hi Yuri,
@satoryboy:
your contribution was well received, and sparked a discussion in this
(https://www.amateurfunk-sulingen.de/forum/index.php?board=19;action=display;threadid=1265;start=msg20093#msg20093)[forum thread]. I would recommend you to join the forum as well or at least read there (it has a translation built in and you can write in english, no problem).

There already a number of extensions to the basic functionality discussed, such as a delay before activation. 
BTW, my last pull request simplified a little your debug menu handling mostly and made it simpler to change the accepted VSWR range. I also restricted the range of valid VSWR threshold values from 1 to 6, this takes away the need for two "OFF" settings. 

In addition I offer my help to get your next pull request to look more clean. A lot of the commits are not really contributing because you were struggling with git. This is for a complex tool like git not uncommon, but makes it even more difficult to follow changes. Your next pull request which looks like this will not be accepted, you have to clean it up (or prevent it from looking like this). Please accept our help on this.
 Max, @m-chichikalov pointed you to our document on the contribution process and also kindly offered to help.  

These rules are there to help us managing the development of not so simple piece of software by keeping the quality and ease of use high enough for our contributors to be able to work with it.

db4ple:
Maybe a very concrete tip:

`git commit -amend`

is a great and easy way to keep the number of commits small when working continously on the same change. I use it a lot in combination with the gitgui from the Git Extensions toolset. The tool makes it easy to add whole files OR only a subset of the changes to a commit in a visual way.
 
There are more ways to do this, see for instance https://www.atlassian.com/git/tutorials/rewriting-history

Most of these don't work with gitgui, you'll need to use the command line (or other tools, I don't use these and don't know which are good at these tasks)
 
Most important rule to follow: never use any of these ideas AFTER a pull request containing the commits you intend to modify got accepted. Git does not have a problem with this technically, but it may create a situation not so easy for you to resolve.

satoryboy:
Thanks for your help. I'm using Git Desktop, in which there is no command to reload the repository. Learning Eclipse IDE...

df8oe:
It is mandatory to use git commands on command line for getting full power out of git. I have not ever used git in a GUI version ( I am running Linux where gits roots are located and these roots where only accessible via command line) but as Danilo already wrote I do not think there is any GUI which is capable to provide the neccessary commands for working effectively with git... So start learning to use git on command line is the only way. Eclipse learning has nothing to do with git! I recommend not to use Eclipse's builtin git functions because they are very rudimentary as all other git GUIs only do have rudimentary and not satisfying git usage.

m-chichikalov:
100% agree with you

df8oe:
Git commands are identical for Linux and Windows (on command line).
Here you can find a very good tutorial:
http://www.vogella.com/tutorials/Git/article.html
Feel free to ask us for support BEFORE you run in trouble.

satoryboy:
Thanks a lot! Never late to learn.

df8oe:
Absolutly not and you only will need a small percentage of git power. So your questions and our answers can be the initial start for a git tutorial at our WIKI - starting with simple handling up to working with only local available branches and rebasing from externalGitHub owned by another person. If git would not be available it must be invented first to do effective community work scattered around the world...

satoryboy:
Yeah, community work for better future. I am absolutely serious.

PR #1657, opened 2018-12-31
Ignoring all files from .settings

m-chichikalov:
Hi @db4ple, take a look, do you really need some specific files in .settings folder? For sure I can add this to the my global .gitignore but I prefer to change here. As long as I'm using other tools in Eclipse there is other files in this folder which I would like to ignore, if it's passible of course.    

Thnk you.


db4ple:
Dear Max, could you please follow the contribution guide lines which asks to use rebase instead of pull/merge.  Saves us one more or less irrelevant commit. Thank you.

I was not closely watching the .gitignore content in this directory. I should have. The one file which at least should be shared is the preferences regarding code formatting. This is relevant in order to ensure similar coding layout. I now saw that this is also excluded, and this is the reason why we have slightly different coding styles in the pull requests.  But anyway, we can handle that with the `!pattern`.
I see no issue here. 

However, there is one more thing. I just checked and we should test if we can get rid of the "language.settings.xml" in the repo which drives me nuts as Eclipse is constantly changing it. At least when I delete the file, it is recreated by Eclipse. We have to see if it still is true that a recent Eclipse needs this file when initially importing the project.

db4ple:
I checked, in Eclipse 2018-09 from the GNU MCU Eclipse project there is no issue, when importing the project into a fresh workspace without the language.settings.xml present. I think we should removed it from the repo, too.

m-chichikalov:
> Dear Max, could you please follow the contribution guide lines which asks to use rebase instead of pull/merge. Saves us one more or less irrelevant commit. Thank you.

Sure, sorry...

db4ple:
Could you please amend your commit and add the remove of the language.setting.xml, then we don't have to create another commit for this, and we simply forget about the merge commit ;-)

m-chichikalov:
> However, there is one more thing. I just checked and we should test if we can get rid of the "language.settings.xml" in the repo which drives me nuts as Eclipse is constantly changing it. At least when I delete the file, it is recreated by Eclipse. We have to see if it still is true that a recent Eclipse needs this file when initially importing the project.

The same on my side )

Let's exclude everything and leave anly needed. I tried to remove "language.settings.xml", the eclipse creates it each time I open project...

so, what if we leave it like this
```
**
!org.eclipse.ltk.core.refactoring.prefs
```
It should make the job, I believe... at least I see that Eclipse save refactoring settings in this file on my side

m-chichikalov:
Ok, maybe I'm not right but looks like when you enable 

you need this guys
```
org.eclipse.cdt.core.prefs
org.eclipse.cdt.ui.prefs
org.eclipse.ltk.core.refactoring.prefs
```

Or we can create .xml with settings for it and ask yourself to make import and leave .gitignore with **

PR #1660, opened 2019-01-02
Improving VSWR Protection (test)

satoryboy:
Just for reduce of codec signal gain, when WSVR Protection is worked.

### DIFF
--- a/mchf-eclipse/drivers/ui/radio_management.c
+++ b/mchf-eclipse/drivers/ui/radio_management.c
@@ -1445,6 +1445,7 @@ bool RadioManagement_UpdatePowerAndVSWR()
         	{
         	RadioManagement_DisablePaBias();
         	swrm.high_vswr_detected = true;
+        	ts.tx_power_factor = 0.001;
         	}
         }
 
--- a/mchf-eclipse/drivers/ui/radio_management.c
+++ b/mchf-eclipse/drivers/ui/radio_management.c
@@ -1445,7 +1445,8 @@ bool RadioManagement_UpdatePowerAndVSWR()
         	{
         	RadioManagement_DisablePaBias();
         	swrm.high_vswr_detected = true;
-        	ts.tx_power_factor = 0.001;
+        	uint8_t tx_band = RadioManagement_GetBand(df.tune_new/TUNE_MULT);
+        	RadioManagement_PowerLevelChange(tx_band, PA_LEVEL_0_5W);
         	}
         }
 


df8oe:
I am not satisfied with this approach of a fixed factor. In my view best would be to clamp output gain fixed to e.g. 0.5W if protection is active. You can use power factors which are used for calculating power on the bands. These are individual for each band. For later use the fallback power can be adjustable (e.g. for use with automatic tuners which need >2W etc.)

For those who are running mcHF with pin diode antenna switching simply cutting off of BIAS is a goal: it reduces oscillation possibility. But in all other cases (relay switching, other PA design etc.) switching is not recommended. Here simple reducing power by factor as described above would be the final solution. So for a first approach reducing power to a defined output power using band specific factors is appreciated.

vy 73
Andreas

db4ple:
What @df8oe wants to say: 
call RadioManagement_SetBandPowerFactor(PA_LEVEL_0_5W) instead of manipulating the value directly.

In general, try not to set variables directly if there is a function to do so. 

If you are using Eclipse, there is an easy way to find out where a variables is used how: click on the variable of interest in the source code, and open the context menu with a right click and select "Referenes->Project". It will show you all read and write accesses to that variable. BTW, same works for functions.

With function key F3 you can jump to the place where something is defined (source code of a function for instance).

sp9bsl:
> With function key F3 you can jump to the place where something is defined (source code of a function for instance).

the same with CTRL+left mouse button on function call or variable/constant name

m-chichikalov:
Hi Юрий, I'm sure Danilo will ask you to squash 2 commits into one, so

Take a look the documentation about interactive rebasing...

Execute this command
**`git rebase --interactive HEAD~2`**  or **`git rebase -i HEAD~2`**
and squash the last two commits together, be aware that commits in opposite order in screen.

The number 2 after HEAD~ means that you are going to interactively rebase two last commits.

If you do something wrong, you can use **`git reflog`** and **`git reset`** to go back at point where the mistake was not done.

After you need to push it with --force flag

And, I know, it's always the pain to choose proper name for commits... I'm sure you would agree that 'May be this way...' in history of changes would be looks ugly.

My problem, that I always do mistake in english words :)

Would you like to learn something new, if YES, I will make comments like this one. If NO, just tell me.

Happy new year.

satoryboy:
YES, of course. Thanks for your help. Sorry, I still use GitHub Desktop for commits.

satoryboy:
I can't rename my last commit. So, what's now?

df8oe:
Have you tried git on command line - search the web for "git commit --amend"?? It is not possible to do that using GUI.

m-chichikalov:
do you have installed "git for windows"? 
There is the [link.](https://gitforwindows.org/)

Try to press **Ctrl+`** in GitDesktop it should open Git BASH...

m-chichikalov:
@satoryboy, you can ask me or others and we will push PR with your changes as one commit if you want. (But it still going to be in your history, so the next time you will do PR from the same branch those commits would be included, you would need to remove this branch and checkout another from main/active-devel)
However, to start working with git CLI is the better option.

satoryboy:
OK, I'm agree. Sorry, I'm in aircraft.
--
Отправлено из Mail.Ru для Android суббота, 05 января 2019г., 18:44 +05:00 от Maksim Chichikalov  notifications@github.com :

satoryboy:
> @satoryboy, you can ask me or others and we will push PR with your changes as one commit if you want. (But it still going to be in your history, so the next time you will do PR from the same branch those commits would be included, you would need to remove this branch and checkout another from main/active-devel)
> However, to start working with git CLI is the better option.

Well, please do it. OM's wait for test VSWR Protection improving. Sorry, I'm far away from my computer, in St. Peterburg. I will use Git for Windows.

m-chichikalov:
Done, you could close this one. 73!

PR #1668, opened 2019-01-02
Update CONTRIBUTING.md

db4ple:
added conditions to "if"

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -93,7 +93,7 @@ No one will be happy with all rules, but as long as the rules are usable, please
   * __Single Line Conditional statements__: please, always use parentheses. C is not Python and indention is easily broken, makes code very hard to debug. 
    
 ```C
-	if 
+	if (c == 7)
 		a = 3;
 	else
 		b = 2;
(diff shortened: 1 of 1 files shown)


m-chichikalov:
I fixed alignments for pictures, will push later.

m-chichikalov:
@db4ple  I also found that there is no mention about wich type of path should be used in `#include`

thsi is from ui_driver.c
``` C
#include <math.h>
#include <src/uhsdr_version.h>
#include "hardware/uhsdr_board_config.h"
#include "ui_lcd_layouts.h"
#include "ui_menu.h"
#include "uhsdr_rtc.h"
#include "adc.h"
#include "drivers/ui/oscillator/osc_si5351a.h"
#include "audio_nr.h"
```
All available variants...

db4ple:
Generally the practice is to just use the name of the include file for local files and no paths.
git blame tells me, @df8oe is not following that style. But you're right, let us write it down. We just should keep in mind: Having more rules doesn't imply that they are being better followed.

There are arguments pro and con. 

Pro: Makes reorganizing code a little easier, helps us to run into problems wiith same name of files earlier but most important, the majority of includes follow this approach.

Con: Makes setting up the correct include path in Make/Eclipse or other build tools more complex. But as we are not adding or removing hundreds of folder day by day, not too much of an issue.

Let us write it down, then:
```
#include "a_header_file.h"
```

PR #1672, opened 2019-01-04
Added enable/disable asserts into release/debug builds.

m-chichikalov:
- Asserts were not working because assert_failed() was empty. Updated to stuck there on the exceptions.

- Added ability to include asserts in build when it's explicitly specified by mention DEBUG=1

### DIFF
--- a/mchf-eclipse/Makefile
+++ b/mchf-eclipse/Makefile
@@ -18,6 +18,8 @@
 # for example to build the bootloader in the build-bl folder inside this directory
 # make -f ../Makefile ROOTLOC=".." bootloader 
 
+# to invoke asserts into the build use make with DEBUG=1; -> make -f "../Makefile" ROOTLOC=".." DEBUG=1 all
+
 # you can set the MCU you want to build for here or via environment variable (F4 or F4-512KB or F7 or H7)
 ifndef BUILDFOR
   BUILDFOR=F4
@@ -30,6 +32,12 @@ ifeq ($(BUILDFOR),F4-512KB)
   CONFIGFLAGS +="-DIS_SMALL_BUILD"
 else
   FW_LINKERFILE_F4=arm-gcc-link_f4_flash1024k.ld
+endif
+
+ifdef DEBUG
+  CONFIGFLAGS += -DDEBUG -DUSE_FULL_ASSERT -DTRACE
+else
+  CONFIGFLAGS += -DNDEBUG
 endif
 
 ifeq ($(BUILDFOR),F7)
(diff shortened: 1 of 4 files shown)


m-chichikalov:
Strange but adding -DNDEBUG is broking 
`The command "make -f ../Makefile OPT_GCC_ARM=$GCC_DIR ROOTLOC=".." BUILDFOR="H7" TRX_ID="i40h7" TRX_NAME="OVI40H7" all" exited with 2.`

Trying understand why.

m-chichikalov:
Strange that 
LIBS := -lm -lc -lnosys
but  --specs=nano.specs  

Should be the same? like --specs=nosys.specs  No?

Seems like some fprintf() invoked when it compiles with NDEBUG for H7. Any thoughts? or _write to errno

df8oe:
This has nothing to do directly with new debug option but with replacing "nano" with "nosys". Due to this the size of the binary is increasing. Not a problem for firmware builds. But bootloader is using nearly the full region of its page (32KB). Slight increasig the size results in build failure (binary is not fitting into 32KB).

db4ple:
In fact, our makefile is kind of wrong in its approach for libraries.  
It turns out, that combining --specs=nano.specs and -lnosys is a neat trick. --specs=nano.specs brings in libc_nano, which expects us to provide implementations for a set of system calls. And -lnosys just contains (mostly dummy) implementations of just that. So in our case, we have to it this way. But I think, after looking into nano.specs and nosys.specs, the right way is to throw out -lc (which brings in the "big" standard libc" and -lnosys and use --specs=nano.specs AND --specs=nosys,specs.

This brings in -lc_nano and -lnosys and produces a binary. I haven't tested it on the device,, but it links and I see no reason why it should not work, since we have been relying on libc_nano.a and libnosys.a before. (the -lc was obviously not used, if we just use --specs=nosys.specs the resulting binary gets larger than before as it now really uses the content of the standard libc.a ). Size of the binary is about as before, so I guess it results in more or less the same binary.

I have to admit, that I did not fully understood the specs approach before looking into the .specs files. You can see there that nosys.specs has a conditional expression depending on the use of nano.specs (i.e. information set in nano.specs).  

So what comes out of this: We can through out -lc and -lnosys and bring in --specs=nosys.specs after (looks like the order is relevant) --specs=nano.specs and we are as good as before.

db4ple:
Turns out, we don't need -lm either. So our LIBS list can be emptied.

m-chichikalov:
> This has nothing to do directly with new debug option but with replacing "nano" with "nosys". Due to this the size of the binary is increasing. Not a problem for firmware builds. But bootloader is using nearly the full region of its page (32KB). Slight increasig the size results in build failure (binary is not fitting into 32KB).

No, this is appears after I tried to change "nano" to "nosys", before the linker crashed because it does not had _write and other system calls.

m-chichikalov:
> I have to admit, that I did not fully understood the specs approach before looking into the .specs files. You can see there that nosys.specs has a conditional expression depending on the use of nano.specs (i.e. information set in nano.specs).

Never looked into specs either, so thanks, quite helpful

m-chichikalov:
Ok, seems that Travis did it successfully, might be merged.
And BTW, the size of binary is 1k less on my side with disabled asserts.

PR #1674, opened 2019-01-05
Review of UHSDR interface for HW interfaces.

m-chichikalov:
- locks.
- uhsdr_hw_spi
- uhsdr_hw_fmc
- uhsdr_structs

@db4ple , @df8oe could you review, do not accept it yet. I pushed it for reviewing because it's very important to choose proper interface to stick with it.

The main advantage is we can slowly go to the point where main aplication code even doesn't know about existing stm32_hal or whatever is in use...

Pay attention to name of functions name of fields in structs and so on... Not everything is done, so...

for example rewriten functions for touch looks like this

``` C
static UHSDR_INTERFACE_t touch_IF;

void UiLcdHy28_TouchscreenInit( uchar mirror )
{
	mchf_touchscreen.xraw = 0;
	mchf_touchscreen.yraw = 0;

	mchf_touchscreen.hr_x = 0x7FFF;                        // invalid position
	mchf_touchscreen.hr_y = 0x7FFF;                        // invalid position

	UHSDR_SPI_GetInstance ( &touch_IF, __UHSDR_SPI2_DISPLAY__ );
	touch_IF.IO->Init( &touch_IF ); // Initiate the HW, in this case SPI

	dio_t cs = { .GPIO_port = TP_CS_PIO, .GPIO_pin = TP_CS };
	UHSDR_SPI_SetChipSelectPin ( &touch_IF, &cs );

	mchf_touchscreen.present = UiLcdHy28_TouchscreenPresenceDetection();
}

static void UiLcdHy28_TouchscreenReadData ( uuint16_t* x_p, uint16_t* y_p )
{
    uchar xpt_response[ XPT2046_COMMAND_LEN ];

    touch_IF.IO->RequestTransaction ( &touch_IF, 100 /*FIXME*/);
    spi_speed_clock_e temp_speed = UHSDR_SPI_ChangeClockSpeed ( &touch_IF, SPI_clock_VerySlow );

    touch_IF.IO->SendReceive_Polling ( &touch_IF, (uint8_t*)xpt2046_command, xpt_response,
            XPT2046_COMMAND_LEN, DATA_8_BIT );

    UHSDR_SPI_ChangeClockSpeed ( &touch_IF, temp_speed );
    touch_IF.IO->CloseTransaction ( &touch_IF );

    *x_p = (xpt_response[ 5 ] << 8 | xpt_response[ 6 ]) >> 3;
    *y_p = (xpt_response[ 3 ] << 8 | xpt_response[ 4 ]) >> 3;
}
```

This code doesn't know about HAL at all, for me it's good sign.

### DIFF
--- a/mchf-eclipse/hardware/uhsdr_hw_fmc.c
+++ b/mchf-eclipse/hardware/uhsdr_hw_fmc.c
@@ -1,0 +1,185 @@
+/*  -*-  mode: c; tab-width: 4; indent-tabs-mode: t; c-basic-offset: 4; coding: utf-8  -*-  */
+/*************************************************************************************
+ **                                                                                 **
+ **                                        UHSDR                                    **
+ **               a powerful firmware for STM32 based SDR transceivers              **
+ **                                                                                 **
+ **---------------------------------------------------------------------------------**
+ **  Description:    FMC boundary. Provides wrappers for HW depended functions.     **
+ **  Last Modified:  rv9yw - m.chichikalov@outlook.com                              **
+ **                  2018-12-21 : Added this file and initial implementation.       **
+ **                                                                                 **
+ **  License:         GNU GPLv3                                                     **
+ ************************************************************************************/
+
+//TODO review dependencies !!!
+#include "uhsdr_board.h"
+#include "uhsdr_mcu.h"
+
+#include <assert.h>
+#include "uhsdr_hw_fmc.h"
+#include "locks.h"
+
+#ifdef UI_BRD_OVI40
+	#include "fmc.h"
+#else
+	#include "fsmc.h"
+#endif
+
+#define DEFAULT_FMC_TIMEOUT 100
+
+static volatile ushort* __address;
+
+// There is only one FMC in use, so implemented in more simple way than SPI.
+static HW_INTERFACE_t available_fmc[ ] = {
+    #if defined ( STM32F4 )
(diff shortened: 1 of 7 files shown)


db4ple:
Well, this is a larger discussion to have.

The main question is: What do we concretely gain from the introduction of this particular abstraction layer and is this gain relevant/of interest for us?

I have a few pros and cons, but I would let you go first.

m-chichikalov:
Ok, lets start from cons
- Slightly increases size of code, but not significantly
- Adds extra layer ( HAL already as a layer)
- maybe a bit complex to understand for not experienced C programmers
- increases number of operations with function pointers, so increased chance to have nullpointer exceptions....( as I'm going write unit tests for this layout it's going to be testable, so not really a cons )
- documentation or guide of use should be written ( Yes, it cons :)

however, this is the price for pros
- One step to break the dependence on the HAL library in main aplication code.  
-- as a result of it much easy write unit tests for modules. ( easy to mock this layer than the whole STM HAL )
-- easy add other MCU ( not stm32 ) -> step to be UniversalHSDR
- As it provide multi thread safe mechanism -> step to easy adopting RTOS
-- no need to put everything in one file everything that works with sharing interface -> reduce dependencies and lead us to more **organised and structured code**.
- provides kind of one level inheritance-abstraction, so main code would be the same for any chosen interface and can be switched in run time.

Probably I missed some cons and prons, but the more **organised and structured code** is the most important I think.

If not start working on the organising and structured the code, the spaghetti ball will only growing and never ever disappear.

PR #1690, opened 2019-01-15
Fixed Sidetone Gain not immediately applied after change / Canary Move / warn refactor

db4ple:
- Now in all modes which have side tone, the change of side tone volume
  is applied immediately (before only in Tune).
- The extracted canary methods are now used.
- added a variable init to keep the compiler silent in freedv_api

### DIFF
--- a/mchf-eclipse/src/uhsdr_main.c
+++ b/mchf-eclipse/src/uhsdr_main.c
@@ -15,7 +15,6 @@
 // Common
 #include "uhsdr_board.h"
 #include <stdio.h>
-#include <malloc.h>
 #include "uhsdr_rtc.h"
 #include "ui_spectrum.h"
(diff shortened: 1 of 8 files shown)


mchf-eclipse/src/uhsdr_main.c
```diff
@@ -46,6 +46,7 @@
 #include "drivers/ui/oscillator/osc_si570.h"
 #include "drivers/audio/codec/codec.h"
 #include "misc/profiling.h"
+#include "uhsdr_canary.h"
```
m-chichikalov, line 48: You could remove including malloc.h up there also, I think...

db4ple: Done.

PR #1693, opened 2019-01-18
 Fixed not properly handled value of Audio PA when TX not configured. - Added sidetone for RTTY and BPSK.

m-chichikalov:
When the PA not configured and tx_power_factor = 0 there is no ANY sidetone in ANY mode. So, I think red square with power indication not enough and this should be handled in some of the ways:

- We do NOT turn on TX at all and write something on the screen that PA is not configured.
- We do turn on TX, play Sidetone (People can practice CW even only with UI).
- ... some other ways, any proposals?

My humble opinion that right now it's not consistent, it's going to TX, no Sidetone... I spent two hours to understand why.

TODO/FIXME :
* [ ] The value of sidetone in RTTY and BPSK can be adjusted only in CW mode, (STG) should be available also in RTTY and BPSK.
* [x] DO not understand part of the code, @db4ple take a look comments below.
* [ ] When pressed Tune in BPSK, not every time, but sometimes the tone of BPSK signal with distortion, do we have the same in TX output?

Looking forward to comments ;)

### DIFF
--- a/mchf-eclipse/drivers/audio/audio_driver.c
+++ b/mchf-eclipse/drivers/audio/audio_driver.c
@@ -4393,6 +4393,23 @@ static void AudioDriver_TxProcessorFM(AudioSample_t * const src, IqSample_t * co
     AudioDriver_TxIqProcessingFinal(FM_MOD_AMPLITUDE_SCALING, swap, dst, blockSize);
 }
 
+#if defined(UI_BRD_OVI40)
+/*
+ * Used only on the OVI40, there is no way on the MCHF to add the sidetone.
+ * should be called the last in the chain.
+ * ( src buffer should be not cleared or tremendously modified by previous stages)
+ */
+static void AudioDriver_FillSideToneAudioBuffer( float32_t* const src, AudioSample_t* const dst,
+        int16_t blockSize, float32_t gain, bool is_signal_active )
+{
+    float32_t final_gain = ( is_signal_active ) ? gain * IQ_BIT_SCALE_UP : 0;
+    for( uint32_t i = 0; i < blockSize; i++ )
+    {
+        dst[i].r = dst[i].l = correctHalfWord((int32_t)src[i] * final_gain);
+    }
+}
+#endif
+
 #ifdef USE_FREEDV
 static void AudioDriver_TxProcessorDigital (AudioSample_t * const src, IqSample_t * const dst, int16_t blockSize)
 {
@@ -4543,25 +4560,29 @@ static inline void AudioDriver_TxFilterAudio(bool do_bandpass, bool do_bass_treb
 }
 
 
-static void AudioDriver_TxProcessorRtty(IqSample_t * const dst, uint16_t blockSize)
+static void AudioDriver_TxProcessorRtty ( IqSample_t * const dst, uint16_t blockSize )
 {
-
-	for (uint16_t idx =0; idx < blockSize; idx++)
-	{
-		adb.a_buffer[0][idx] = Rtty_Modulator_GenSample();
-	}
-    AudioDriver_TxFilterAudio(true,false, adb.a_buffer[0], adb.a_buffer[0], blockSize);
-    AudioDriver_TxProcessorModulatorSSB(dst, blockSize, ts.iq_freq_mode, ts.digi_lsb);
-
-    // remove noise if no CW is keyed
-    memset(adb.a_buffer[0],0,sizeof(adb.a_buffer[0][0])*blockSize);
-
-    if (ts.cw_keyer_mode != CW_KEYER_MODE_STRAIGHT)
+    for ( uint16_t idx = 0; idx < blockSize; idx++ )
     {
-    	CwGen_Process(adb.a_buffer[0], adb.a_buffer[0], blockSize);
-    	// we just misuse adb.a_buffer[0], and generate a CW side tone
+        adb.a_buffer[0][idx] = Rtty_Modulator_GenSample( );
     }
+    AudioDriver_TxFilterAudio ( true, false, adb.a_buffer[0], adb.a_buffer[0],
+            blockSize );
+    AudioDriver_TxProcessorModulatorSSB ( dst, blockSize, ts.iq_freq_mode, ts.digi_lsb );
 
+    if ( ts.cw_keyer_mode != CW_KEYER_MODE_STRAIGHT )
+    {
+        // remove noise if no CW is keyed
+        memset ( adb.a_buffer[1], 0, sizeof( adb.a_buffer[1][0]) * blockSize );
+        CwGen_Process ( adb.a_buffer[1], adb.a_buffer[1], blockSize );
+
+        #if defined(UI_BRD_OVI40)
+        // FIXME if we leave two signals together, level of RTTY should be adjusted somehow.
+        // now it's scaled down to really low level.
+        arm_scale_f32 ( adb.a_buffer[0], 0.05, adb.a_buffer[0], blockSize );
+        arm_add_f32 ( adb.a_buffer[0], adb.a_buffer[1], adb.a_buffer[0], blockSize );
+        #endif // UI_BRD_OVI40
+    }
 }
 
 static void AudioDriver_TxProcessorPsk(IqSample_t * const dst, uint16_t blockSize)
@@ -4574,7 +4595,9 @@ static void AudioDriver_TxProcessorPsk(IqSample_t * const dst, uint16_t blockSiz
 
     AudioDriver_TxFilterAudio(true,false, adb.a_buffer[0], adb.a_buffer[0], blockSize);
 	AudioDriver_TxProcessorModulatorSSB(dst, blockSize, false, false);
+
 /*
+	// TODO There is a bug when CW enabled in BPSK, somehow affected the AGC struct and it crashes.
     memset(adb.a_buffer[0],0,sizeof(adb.a_buffer[0])*blockSize);
 
     if (ts.cw_keyer_mode != CW_KEYER_MODE_STRAIGHT)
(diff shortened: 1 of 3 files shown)


mchf-eclipse/drivers/audio/audio_driver.c
```diff
@@ -4553,14 +4571,14 @@ static void AudioDriver_TxProcessorRtty(IqSample_t * const dst, uint16_t blockSi
...
-    if (ts.cw_keyer_mode != CW_KEYER_MODE_STRAIGHT)
-    {
-    	CwGen_Process(adb.a_buffer[0], adb.a_buffer[0], blockSize);
-    	// we just misuse adb.a_buffer[0], and generate a CW side tone
-    }
+//    if (ts.cw_keyer_mode != CW_KEYER_MODE_STRAIGHT)
+//    {
+//    	CwGen_Process(adb.a_buffer[0], adb.a_buffer[0], blockSize);
+//    	// we just misuse adb.a_buffer[0], and generate a CW side tone
+//    }
```
m-chichikalov: @db4ple , Danilo, could you explain, please. Did not understand why we called CW processing in RTTY. 
To input characters from CW key and transmit it as RTTY? or what?

db4ple: Yes, exactly. And for that reason, this change should not be accepted (as it would break this function).
Please remove from pull request.

db4ple:
Well, if one follows the instructions to calibrate the PA before pressing TX for real, there is a sidetone at least in CW. Once calibrated, you can practice CW even with TX disabled (long press on Tune to disable TX) and get a sidetone. 

Ok, I agree, not everyone strictly follows the instructions. I have heard of such cases, of course I would never ever do such a thing.

To get out of this problem:
We could generate a warning at boot, but must not prevent TX as you need this to calibrate the TRX.

Otherwise you have a catch 22 situation: You want to calibrate TX,  to do that you have turn on TX, but this doesn't work, because the TX power factor is 0 and changing is not possible until you have turned on the TX.

BTW, I am currently working on restructuring the audio_driver and also came across the sidetone. I was thinking to do the reverse (letting CW produce sidetone audio into the a_buffer) and then keep the actual sidetone "production" at its original location. Produces less "#ifdef". But lets go for your idea.

I think, as you wrote, we should leverage the existing sidetone gain setting also for the OVI40 sidetones.

As for the restructuring work: See here: https://github.com/db4ple/UHSDR/tree/audio_driver_breakup
Goal is to better understand how we can decouple iq sample rate from audio sample rate and to prepare this. Of course, part of this is also to make the code easier to read and maintain (as far as this is possible).

mchf-eclipse/drivers/audio/audio_driver.c
m-chichikalov: Ok, I got it.
In this case, maybe leave it but combine signals, RTTY and CW tone, make RTTYvolume less than CW. 
Or make it adjustable. I mean if someone want to hear the only rtty and do not care about "cw way entering" characters? What do you think?

m-chichikalov:
> I think, as you wrote, we should leverage the existing sidetone gain setting also for the OVI40 sidetones.
> 

The existing sidetone setting affects every sidetone which I add, but the problem it's available to adjust on the screen only in CW mode. I did not look, probably in UI driver somewhere should be allowed to change it in others modes.

df8oe:
Everyone who enters characters via paddle (regardless what mode) wants to listen to sidetone. It is impossible to key in characters without hearing them. And I think level of sidetone is adjusted to a comfortable individual level which matches all modes where keying in via paddles is possible, too.

m-chichikalov:
> Everyone who enters characters via paddle (regardless what mode) wants to listen to sidetone. It is impossible to key in characters without hearing them. And I think level of sidetone is adjusted to a comfortable individual level which matches all modes where keying in via paddles is possible, too.

Yes, it's working as you discribed. 

But let's decide do we want to combine into lineout two signals RTTY and CW,  This is my personal preference but I would like to hear both when I enter text by paddles.

m-chichikalov:
And seems like that OVI40 has two channels (left and right) not uses enough, we can put cw to right channel rtty to left...

db4ple:
Using both channels: We cannot assume that the audio speaker is connected to both channels (at least in my case it is only on the left. So Iwould be able to hear only one channel). Historically I think the left channel is the main channel but essentially unless there is an explicity stereo use case, we should put the same audio on both channels (or less information on the right channel, i.e. the unmodified signal).

db4ple:
BTW, you can adjust sidetone in all modes which support a sidetone, see here:
https://github.com/df8oe/UHSDR/blob/be49158b652d91afadf46b5f47a3a1ddd085165a/mchf-eclipse/drivers/ui/ui_driver.c#L3395

I admit that this change is fairly recent ;-)

m-chichikalov:
Ok, let's try mix RTTY and CW, and will react on the users' feedback. If users say "It's terrible" I will quickly revert it. No one will know until we try, do you agree?

@db4ple , Danilo, why the CW processing not called in BPSK?  I see it's commented out.

db4ple:
Hi,
why not enabeld in BPSK? No idea. Wasn't my commit. I did not notice until recently that it is off.

m-chichikalov:
> Hi,
> why not enabeld in BPSK? No idea. Wasn't my commit. I did not notice until recently that it is off.

Ok, I will try to enable and do the same as RTTY.

mchf-eclipse/drivers/audio/audio_driver.c
```diff
@@ -4670,10 +4696,14 @@ static void AudioDriver_TxProcessor(AudioSample_t * const srcCodec, IqSample_t *
 
         if (signal_active)
         {
+    #if defined(UI_BRD_OVI40)
+            // FIXME -> there is hard coded gain, should be defined or be adjustable.
+            AudioDriver_FillSideToneAudioBuffer( adb.i_buffer, audioDst, blockSize, 0.1 );
+    #endif
```
db4ple: I am not sure this works as expected. CW does not have signal active all the time and then the audio buffer contains whatever was there last time. In the original place we mute the signal in dst AND then copy it to audio, so there is no problem.

mchf-eclipse/drivers/audio/audio_driver.c
```diff
@@ -4777,28 +4807,35 @@ static void AudioDriver_TxProcessor(AudioSample_t * const srcCodec, IqSample_t *
...
-    if ((void*)audioDst != (void*)dst)
-    {
-    	if (dmod_mode == DEMOD_CW)
-    	{
-    		for (uint16_t idx = 0; idx < blockSize; idx++)
-    		{
-    			audioDst[idx].r = audioDst[idx].l = dst[idx].l;
-    		}
-    	} else if (is_demod_rtty())
-    	{
```
db4ple, line 4792: If you would replace  if(is_demod_rtty()) with if(RadioManagement_UsesTxSidetone())  this should almost do the same job as your code.

mchf-eclipse/drivers/audio/audio_driver.c
m-chichikalov: I see what do you mean, so far working fine, clean and nice sidetone. To be save changed a bit, take a look...

mchf-eclipse/drivers/audio/audio_driver.c
m-chichikalov: For RTTY and BPSK but not for CW, but it's a good idea, thanks I will think about it.
ts.tx_power_factor shouldn't be used for audio output.

m-chichikalov: Maybe change CW to use a_buffer for generating sine, so all modes will have one source of signal for audio?

m-chichikalov: Are you going to change AudioDriver_TxIqProcessingFinal() that it will receive source as a parameter? 

Another notice, we have TX_Audio_via_USB in "Debug settings" -> Source, Filtered and Final IQ, so, in many modes the source buffer modified during processing. Maybe we need dedicate fixed buffer for the source,  What's a difference between Filtered and final IQ?

db4ple: >
> 
> Maybe change CW to use a_buffer for generating sine, so all modes will have one source of signal for audio?

I was thinking of this.



> 
> 
> Are you going to change AudioDriver_TxIqProcessingFinal() that it will receive source as a parameter?
>

Yes, finally.
 
> Another notice, we have TX_Audio_via_USB in "Debug settings" -> Source, Filtered and Final IQ, so, in many modes the source buffer modified during processing. Maybe we need dedicate fixed buffer for the source, What's a difference between Filtered and final IQ?

It is not filtered IQ, it is filtered audio input (!) after filtering and compression. It is in fact only FM where the a_buffer is being "destroyed" and if we really want, we could use a_buffer[1] for that, so that a_buffer[0] remains intact.

m-chichikalov: > > Are you going to change AudioDriver_TxIqProcessingFinal() that it will receive source as a parameter?
> 
> Yes, finally.
> 
Ok, in this case, I would like to change buffer for generated CW wave to a_buffer and add HACK into 
AudioDriver_TxIqProcessingFinal() for this and later me or you change when parameter would be added. 

That gives to put filling audio buffer in one place is it was before.

db4ple: I don't see why you have to  change AudioDriver_TxIqProcessingFinal() , you can simply copy the i_buffer (and scale a_buffer) before the AudioDriver_TxIqProcessingFinal is being run. In case of signal_active == false you have to fill the a_buffer probably with silence.

m-chichikalov:
Found why in PSK CW processing was commented, when it's active where some nasty bug, when switching from TX to RX, somehow fields in agc struct corrupted and it stucks in exception irq.

Will leave it commented out for now.

mchf-eclipse/drivers/audio/audio_driver.c
m-chichikalov: One or another, doesn't matter.. Ok.

PR #1694, opened 2019-01-19
Added default UI and RF boards defines to handle make all.

m-chichikalov:


df8oe:
Why not going this way:

ifndef CONFIGFLAGS
  CONFIGFLAGS=-DUI_BRD_MCHF -DRF_BRD_MCHF
endif

??

m-chichikalov:
It's almost the same as the assignment to CONFIGFLAGS ignored by make if this parameter passed as an argument. But, sure I can add, maybe it's more clear to read when there this extra "ifndef"...

df8oe:
...it's more clear to read when there this extra "ifndef"...

Yes, that's my opinion, too...

m-chichikalov:
Done.

m-chichikalov:
Wait, there is some error in the building, do not accept. I need to go, will fix later.

m-chichikalov:
@df8oe , Andreas, you can accept, I've just built it in different branch :)

PR #1695, opened 2019-01-19
Tx Processor separation, Issue fixes, Embitz "news"

db4ple:
This is a huge change, unfortunately I had to reapply some of  changes from #1693 manually. I was not able to do this in a nicer way. However, changes still seem to be working. Although I had to adapt them to my changed structure.  I am sorry for  that.

- Embitz marked as unmaintained now (I made the bootloader for the mcHF build, but nothing else), Readme tells it now.
- Moved the I2C access out of the interrupt handler (should not be done there)
- Moved the transmit data processing out of the audio driver and cleaned it up. It should work as before,
  as now also BPSK can use the CW as input  source, RTTY now will be able to generate a sidetone on the mcHF, too (since we are transmitting at zero if (like PSK, CW and FreeDV).
- Cleaned up a lot of complicated things in terms of tx signal processing, especially the USB vs. LSB vs. freq shift things are now much better to understand and clearly separated. 
- Ultimate goal is to make RX/TX sample rates independent from IQ sample rates so that we can switch to 96k or higher.

For more information, please see the commit messages.


m-chichikalov:
Hi, Danilo, as I wrote, I found that it crashed when I tried to enable CW input for BPSK. Did you see the same and found why? Or nothing noticed?

db4ple:
No, it did not crash. It may have to do with my other changes. Although I would not know why.

To be honest, I did not test PSK reception on the other end, but the waveforms looked and sounded ok. But I was able to go on and off TX just like for RTTY.

mchf-eclipse/drivers/audio/audio_agc.c
```diff
@@ -0,0 +1,567 @@
...
+#endif
+        }
+        return;
+    }
+
+    for (uint16_t i = 0; i < blockSize; i++)
+    {
+        if (++agc_wdsp.out_index >= agc_wdsp.ring_buffsize)
+        {
+            agc_wdsp.out_index -= agc_wdsp.ring_buffsize;
```
m-chichikalov, line 333: There is a member field ring_buffsize. Seems like it is constant, init and never change later. Maybe get rid of it and use AGC_WDSP_RB_SIZE instead?

db4ple: I don't know enough about the WDSP AGC to tell you if there is a reason for changing the size dynamically at some point. You have to ask @DD4WH for more info on this code or checkout the original WDSP library.

m-chichikalov: Ok, thanks.

m-chichikalov:
> No, it did not crash. It may have to do with my other changes. Although I would not know why.
> 
> To be honest, I did not test PSK reception on the other end, but the waveforms looked and sounded ok. But I was able to go on and off TX just like for RTTY.

Ok it's building right now, I mean your branch, I'll try and let you know, maybe it's because of power_factor still 0 on my side...

m-chichikalov:
Tried BPSK, it works, no crashes.
But I again do not hear anything not when I press Tune, or when TX in CW,RTTY and BPSK.

db4ple:
No sidetone on a mcHF or an OVI40?

m-chichikalov:
OVI40

db4ple:
I think, I know why. I removed correctHalfWord in the Tx Sidetone Fill function after I moved the TxIqFinalProcessing down. But this is of course wrong, since audioDst != dst. And if both IQ and audio codec are using the 32bit format, we need it for both.  
EDIT: But not on a OVI40. Here we need to shift the AUDIO by 16 bits. And I removed that code, too.
Sometimes it is really difficult...

db4ple:
Ok, now the sidetone is back. I find the mixture between CW "feedback" and the generated signal difficult to manage. We'll probably have to be a little smarter here. But anyway, you'll be able to hear it again.

m-chichikalov:
> Ok, now the sidetone is back. I find the mixture between CW "feedback" and the generated signal difficult to manage. We'll probably have to be a little smarter here. But anyway, you'll be able to hear it again.

Do you want me to check? if so, will do it tomorrow.

PR #1698, opened 2019-01-20
 Fixed BPSK TX, CW input for RTTY and BPSK,  TxProcessor cleanup

db4ple:
Fixed BPSK TX, CW input for RTTY and BPSK

- RTTY and BPSK now accept input from USB keyboard AND morse paddles (if
  an keyer mode is set, not for Straight Key).

- now the BPSK transmit works (at least fldigi is able to decode)
  I did not try the receiving part, but it is supposed to work.
- reduced BPSK gain a little.
- fixed a minor thing in the softdds. It now starts the sine with the first
  value (which is 0) and not the next increment (which is a litte above zero).

Note: I found in cw_gen.c the beginning of a keyboard to morse converter.
This was eating RTTY/BPSK keyboard input. It seems not to be operational
since it requires the TRX in TX mode to accept input, but in CW the starts
only once you use a paddle...

TxProcessor cleanup

- more comments
- some more streamlining of the various modulators
- properly handled filter (re-)configuration whith the tx processor
  being actively run.

What is missing still is the relocation and renaming of functions in source
and header files but we'll postpone this until we are sure this whole
thing works as intended.

### DIFF
--- a/mchf-eclipse/drivers/audio/cw/cw_gen.c
+++ b/mchf-eclipse/drivers/audio/cw/cw_gen.c
@@ -579,7 +579,10 @@ static void CwGen_CheckKeyerState(void)
 
 	if (!ts.cw_text_entry && ps.cw_state == CW_IDLE)
 	{
-		if (ts.txrx_mode == TRX_MODE_TX && DigiModes_TxBufferHasData() && !ps.sending_char &&
+	    // FIXME: I am not sure how this is supposed to work, but without check for DEMOD_CW it happily eats
+	    // all keyboard input for modes such as RTTY or PSK...
+	    // it seems to be intended for transmitting morse via keyboard input...
+		if (ts.dmod_mode == DEMOD_CW && ts.txrx_mode == TRX_MODE_TX && DigiModes_TxBufferHasData() && !ps.sending_char &&
 				(! (ps.port_state & CW_END_PROC) && ps.space_timer < ps.space_time - ps.dah_time))
 		{
 			DigiModes_TxBufferRemove(&c);
@@ -809,7 +812,7 @@ void CwGen_AddChar(ulong code)
 
 			// FIXME: HACK HACK FOR RTTY TX TESTING
 			// We can use the same buffer for all digimodes
-			if ((ts.txrx_mode == TRX_MODE_TX && is_demod_rtty()) || ts.cw_text_entry) // Can rtty rely just on cw_text_entry here?
+			if ((ts.txrx_mode == TRX_MODE_TX && (is_demod_rtty() || is_demod_psk())) || ts.cw_text_entry) // Can rtty rely just on cw_text_entry here?
 			{
 				DigiModes_TxBufferPutChar(result);
 			}
--- a/mchf-eclipse/drivers/audio/psk.c
+++ b/mchf-eclipse/drivers/audio/psk.c
@@ -225,7 +225,9 @@ psk_ctrl_t psk_ctrl_config =
 PskState  psk_state;
 
 void Bpsk_ResetWin() {
-	softdds_setFreqDDS(&psk_bit_dds, psk_speeds[psk_ctrl_config.speed_idx].value / 2, ts.samp_rate, true);
+    // little trick, we just reset the acc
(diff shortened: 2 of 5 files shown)


mchf-eclipse/drivers/audio/cw/cw_gen.c
```diff
@@ -579,7 +579,10 @@ static void CwGen_CheckKeyerState(void)
 
 	if (!ts.cw_text_entry && ps.cw_state == CW_IDLE)
 	{
-		if (ts.txrx_mode == TRX_MODE_TX && DigiModes_TxBufferHasData() && !ps.sending_char &&
+	    // FIXME: I am not sure how this is supposed to work, but without check for DEMOD_CW it happily eats
+	    // all keyboard input for modes such as RTTY or PSK...
+	    // it seems to be intended for transmitting morse via keyboard input...
+		if (ts.dmod_mode == DEMOD_CW && ts.txrx_mode == TRX_MODE_TX && DigiModes_TxBufferHasData() && !ps.sending_char &&
 				(! (ps.port_state & CW_END_PROC) && ps.space_timer < ps.space_time - ps.dah_time))
```
m-chichikalov, line 586: Hi Danilo, I handled already it, and it would be nice if you checkout this file back. It would be much easy for me to rebase later. Thank you...

m-chichikalov:
Nice job regarding BPSK...
One moment, maybe you will fix it in this PR...
Right now the GenCw.c is depends on the state of TRX, it's checking all the time is it in TX or RX or it's TEXT_ENTRY... this is not right, do you agree? Lets change a bit TxProccesing, and call CwProcess only when TRX in TX state, do you see what I mean? The CwGen should not care about TRX state, if someone call it, it should generate CW...

I know that it was done to handle glitch at the start of first dit or dot because Tx proccesing start calling it much early than TRX in TX....

m-chichikalov:
> Nice job regarding BPSK...
> One moment, maybe you will fix it in this PR...
> Gight now the GenCw.c is depends on the state of TRX, it's checking all the time is it in TX or RX or it's TEXT_ENTRY... this is not right, do you agree? Lets change a bit TxProccesing, and call CwProcess only when TRX in TX state, do you see what I mean? The CwGen should not care about TRX state, if someone call it, it should generate CW...

I know that it was done to handle glitch at the start of first dit or dot because Tx proccesing start calling it much early than TRX in TX....

mchf-eclipse/drivers/audio/cw/cw_gen.c
m-chichikalov: If, it's a problem don't worry I will do it, not a big deal

db4ple:
I am not sure what you mean with "it is checking all the time". The CwGen Processing is only called from interrupts.  The call in the rx path comes from @phaethon , I am not sure why exactly this is there. 
I think we should leave it for now and analyze it carefully.

mchf-eclipse/drivers/audio/cw/cw_gen.c
db4ple: What did you handle? This was the fix so that CwGenProcess all keyboard input from RTTY and PSK.
I'll break this out into a single commit, so that you can skip or easily revert it if necessary

m-chichikalov:
> I am not sure what you mean with "it is checking all the time". The CwGen Processing is only called from interrupts. The call in the rx path comes from @phaethon , I am not sure why exactly this is there.
> I think we should leave it for now and analyze it carefully.

Sure, take a look, but I think we should remove this kind of dependencies from GenCw... It should not care about stuff like this... It should do only one work, generate Cw when someone call it.

Thanks

mchf-eclipse/drivers/audio/cw/cw_gen.c
m-chichikalov: > What did you handle? 

I implemented kind of multiplexer for digi_buffer, so only consumer which was set can remove from buffer, also only digi_buffer send character to UI_buffer on input or on outputing. It gives chance to broke infinite loop when CW produces char -> put it into buffer -> and remove the same input from buffer.

PR #1699, opened 2019-01-21
Fixed hardly modulated by CW RTTY and BPSK sidetone.

m-chichikalov:
Danilo, because of the fact that in your implementation the tone mixing and scaled only when CW is active there is nasty tone hardly modulated by CW in the headphone. Try this one, it's clean and nice tone.

So, another notice, we have to destination for the filtered source, headphone and USB, seems like the RTTY and BPSK tone with CW modulation is useless in USB, but we cannot enter by paddles without CW tone in Headphones, maybe leave RTTY and BPSK not modified for USB and modify and mix with CW for headphones only?


df8oe:
Headphones for OVI40 is same as speaker. Only mcHF has different output for headphone and speaker.

PR #1702, opened 2019-01-21
Refactored/Fixed/Added inputing by CW paddles & USB keyboard & F-butt…

m-chichikalov:
- Fixed issue with multi-button pressing on the keyboard.
- Added functionality into digi_buffer.
- To record F-buttons could be used Keyboard as well as CW.
- UI F-buttons assigned to F5,F6 and F7 on the keyboard.
- Digi_buffer might be reset by pressing ESC.
- UI buffer might be cleaned by pressing Backspace.

TODO:
- [ ] - Fix CW sending prosings from digi buffer. Seems it was done not right.  
- [ ] - Edit Iambic key state machine to use it for straight key.  (deprecate StraightKey handler)
- [ ] - Rename CW folder into digi_modes and move there rtty and psk modules.
- [ ] - ( need discussion) Two buffers, digi and ui can be combined into one, this gives us a chance to implement features like:
-- Display in different colors what was entered into the buffer and what was already transmitted.
--  handle "backspace" from keyboard to edit what was entered and not transmitted yet..
-- reduce size ;) and some others which I do not see now.

### DIFF
--- a/mchf-eclipse/drivers/audio/psk.c
+++ b/mchf-eclipse/drivers/audio/psk.c
@@ -13,6 +13,7 @@
 #include "radio_management.h"
 #include "rtty.h"
 #include <stdlib.h>
+#include "uhsdr_digi_buffer.h"
 
 // table courtesy of fldigi pskvaricode.cxx
 static uint16_t psk_varicode[] = {
@@ -620,11 +621,11 @@ int16_t Psk_Modulator_GenSample()
 			    // normal characters don't have 2 zeros following each other
 				psk_state.tx_zeros++;
 			}
-			else if (DigiModes_TxBufferRemove(&psk_state.tx_char))
+			else if ( DigiModes_TxBufferRemove( &psk_state.tx_char, BPSK ))
 			{
 			    // if all zeros have been sent, look for new
 			    // input from input buffer
-				psk_state.tx_bits = Bpsk_FindCharReversed(psk_state.tx_char);
+				psk_state.tx_bits = Bpsk_FindCharReversed( psk_state.tx_char );
 				// reset counter for trailing/leading zeros
 				psk_state.tx_zeros = 0;
 				// reset counter for trailing zeros
--- a/mchf-eclipse/drivers/ui/ui_driver.c
+++ b/mchf-eclipse/drivers/ui/ui_driver.c
@@ -51,6 +51,7 @@
 #include "config_storage.h"
 
 #include "cw_gen.h"
+#include "uhsdr_digi_buffer.h"
 
 #include "radio_management.h"
 #include "soft_tcxo.h"
@@ -535,6 +536,7 @@ void UiDriver_TextMsgClear()
 	}
 	ui_txt_msg_buffer[fillcnt]='\0';
 
+	// FIXME there is more affective way to clear space on the screen.
     UiLcdHy28_PrintText(ts.Layout->TextMsgLine.x,ts.Layout->TextMsgLine.y, ui_txt_msg_buffer,Yellow,Black,ts.Layout->TextMsg_font);
     ui_txt_msg_idx = 0;
     ui_txt_msg_update = true;
@@ -3039,6 +3041,7 @@ void UiDriver_SetDemodMode(uint8_t new_mode)
 	};
 #endif
 
+	DigiModes_TxBufferReset();
 	switch(ts.dmod_mode)
 	{
 	case DEMOD_DIGI:
@@ -3046,6 +3049,7 @@ void UiDriver_SetDemodMode(uint8_t new_mode)
 		switch(ts.digital_mode)
 		{
 		case DigitalMode_RTTY:
+		    DigiModes_Set_BufferConsumer( RTTY );
 			if (ts.enc_one_mode != ENC_ONE_MODE_AUDIO_GAIN)
 			{
 				ts.enc_one_mode = ENC_ONE_MODE_RTTY_SPEED;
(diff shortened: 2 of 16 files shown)


db4ple:
Good job! But one serious issue: You are modifying STM_HAL files. This is not recommended, as it impairs our ability to update the STM_HAL easily. I know the change is small but can we do it in a different way?

m-chichikalov:
I know, but there is not another way, one way make code redundant by copying look up tables from stm files to somewhere to use them))) Your proposal? Btw, I will take a look later, maybe figure out something...

mchf-eclipse/drivers/ui/ui_driver.c
```diff
@@ -6940,6 +6960,82 @@ void UiDriver_TaskHandler_HighPrioTasks()
 
 }
 
+static void UiDriver_HandleUSB_Keyboard()
+{
+#ifdef USE_USBKEYBOARD
+    HID_KEYBD_Info_TypeDef *k_pinfo;
+    static uint8_t keys_buffer[sizeof(k_pinfo->keys)] = { 0 };
+
```
db4ple: Even if the data structure is of type uint8_t is is recommended to use the sizeof(k_pinfo->keys)/sizeof(*k_pinfo->keys)
for the number of elements.
Since you are using it multiple times, why not use a constant with an appropriate name?

mchf-eclipse/drivers/ui/ui_driver.c
```diff
@@ -6940,6 +6960,82 @@ void UiDriver_TaskHandler_HighPrioTasks()
...
+                    if ( keys_buffer[i] == k_pinfo->keys[idx])
+                    {
+                        is_exist_in_buffer = true;
+                    }
+                }
+                if ( is_exist_in_buffer == true )
+                {
+                    continue;
+                }
+                else
```
db4ple, line 7009: The code above is a bit of a mystery to me. Please comment. I can see what it does but have no idea what it is supposed to achieve (at least I think so: whenever a key in the newly reported  structure is also in the last set of keys, ignore it.) But why? The data is read from a fifo filled by the USB host controller code. I would not assume we see the same key twice (but we see of course repeated presses). 
At least for me there has never been an issue with "bouncing" keys.

BTW, for difficult to understand or not obvious code, leaving a comment for the other readers of the code is helpful.

db4ple (changes requested):

mchf-eclipse/drivers/ui/ui_driver.c
m-chichikalov: > At least for me there has never been an issue with "bouncing" keys.

it's not "bouncing"
As I asked, try to press two buttons at the same time, what do you see?
Or hold one and press another while you holding first one. 

> The data is read from a fifo filled by the USB host controller code. 

Maybe keyboards different but according to HID specification it's not FIFO

order is not determinated and this is responsobility of the host to figure out... 

It could be done more elegant if order was deterministic, but... save to fallow HID specification I think.

Sure, I will comment later...

mchf-eclipse/drivers/ui/ui_driver.c
m-chichikalov: > Good job! But one serious issue: You are modifying STM_HAL files. This is not recommended, as it impairs our ability to update the STM_HAL easily. I know the change is small but can we do it in a different way?

Did you find some way to not modify STM stuff? I didn't, maybe put USB handling into uhsdr_usb_kb.c, copy stuff from stm where? anyway looks like handling USB will grow, I would like to add some more features, change CW speed, change band, maybe mode, recording F-bottuns on the long press, maybe users suggest something more.... @db4ple , What do you think?

db4ple: Yes, I think I do. I thought I already posted it. We have two options:
Reimplemented the usbh_keybd.c code (one for all platforms, we did that for our composite usb device already).

Or: We simply copy the character of interest to the first location in the array and call the unmodified function. Which is simpler and would do the job for now. 

General remark: Keyboards are a pain because the USB spec did not specific a keyboard layout identification. It feels so much eigthies to have to select the proper layout manually...

mchf-eclipse/drivers/ui/ui_driver.c
db4ple: I see. What your code does is the following: If I press a `M` and keep it pressed, and then press an `A` your code would generate a M and then an A as output. But I thought this is wrong. However, I tried it on my PC and yes, it does generate n M and then an A, so your code is not wrong but I was. Never noticed this. 

However, I still think we should point the reader of this code into the right direction with a comment. Otherwise it is difficult to understand what is the purpose of this piece of code.

mchf-eclipse/drivers/audio/psk.c
```diff
@@ -13,6 +13,7 @@
 #include "radio_management.h"
 #include "rtty.h"
 #include <stdlib.h>
```
db4ple, line 15: We can remove rtty.h now.

mchf-eclipse/drivers/ui/ui_driver.c
m-chichikalov: > Yes, I think I do. I thought I already posted it. We have two options:
> Reimplemented the usbh_keybd.c code (one for all platforms, we did that for our composite usb device already).
> 
> Or: We simply copy the character of interest to the first location in the array and call the unmodified function. Which is simpler and would do the job for now.
> 
> General remark: Keyboards are a pain because the USB spec did not specific a keyboard layout identification. It feels so much eigthies to have to select the proper layout manually...

The second one is a mess. I will take a look at the first one....

PR #1703, opened 2019-01-21
Fixed check for wrong samplerate (which disabled BPSK/RTTY/CW decoder)

db4ple:
- a wrong check effectively disabled the decoders


m-chichikalov:
@db4ple , do you hear that tone of bpsk is different on each switch rx->tx->rx->tx? And very often it's not a good tone? Not sure that this an issue... Just mention.

PR #1704, opened 2019-01-22
Refactored/Fixed/Added inputing by CW paddles & USB keyboard & F-butt…

m-chichikalov:
The previous discussion was [here](https://github.com/df8oe/UHSDR/pull/1702)
 
- Adopted Eclipse/Make settings to fallow new HID driver.
- Fixed issue with multi-button pressing on the keyboard.
- Added functionality into digi_buffer.
- To record F-buttons could be used Keyboard as well as CW.
- UI F-buttons assigned to F5,F6 and F7 on the keyboard.
- Digi_buffer might be reset by pressing ESC.
- UI buffer might be cleaned by pressing Backspace.

TODO:
- [ ] - Fix CW sending prosings from digi buffer. Seems it was done not right.  
- [ ] - Edit Iambic key state machine to use it for straight key.  (deprecate StraightKey handler)
- [ ] - Rename CW folder into digi_modes and move there rtty and psk modules.
- [ ] - ( need discussion) Two buffers, digi and ui can be combined into one, this gives us a chance to implement features like:
-- Display in different colors what was entered into the buffer and what was already transmitted.
--  handle "backspace" from keyboard to edit what was entered and not transmitted yet..
-- reduce size ;) and some others which I do not see now.


db4ple:
@df8oe : Please do not release after merge, Eclipse will not build this correctly as it will automatically compile the "old" and the "new" usb driver files and will end up with multiple defined symbols. The solution is to remove the problematic files from the source code and do a little more homework. I will do this. 

@m-chichikalov: I had a simpler solution in mind, in which we only provide a new function building on top of the existing (and left in place driver) of which we use the basic hid functions but here we go.

df8oe:
@db4ple : Sorry . I have tested this before merging and release build process has started immediately after merging automatically...

db4ple:
Well, then so be it, I well finish my PR soon. Now already testing it. Not a big deal.

m-chichikalov:
> @df8oe : Please do not release after merge, Eclipse will not build this correctly as it will automatically compile the "old" and the "new" usb driver files and will end up with multiple defined symbols. The solution is to remove the problematic files from the source code and do a little more homework. I will do this.
> 
> @m-chichikalov: I had a simpler solution in mind, in which we only provide a new function building on top of the existing (and left in place driver) of which we use the basic hid functions but here we go.

Danilo, Thank you that you corrected (seems like I removed old one from the list of sources, strange). I would change myself, but I had been sleeping, sorry ))) 
And the simpler not always good one.

db4ple:
Eclipse does include all sources in the defined source tree folder lists unless you exclude specific files or folders. And Eclipse does not use the UHSDR makefiles at all.

m-chichikalov:
> Eclipse does include all sources in the defined source tree folder lists unless you exclude specific files or folders. And Eclipse does not use the UHSDR makefiles at all.

I know, I added these old ST folders to excluded filters, strange... I will take a look at your change, I missed something. BTW, I need to fix Building process in Eclipse do be able to test on my side. I would mention that did not test Eclipse settings.

db4ple:
Since last time you were not able to handle Eclipse and I overlooked .cproject as changed, I was not aware that you tested and fixed this already. And because of this I did not test,  I know what Eclipse would does if it is not handled.  Now I will remember you also check the Eclipse builds. Very good.

Nevertheless, I try to limit the number of exclusion in the source try, since I don't find this mechanism a little difficult to manage for multiple different build targets in Eclipse. Or do you know an easy way to list all the build exclusions for all projects in some kind of readable format?

Anyway, this better aligned all three devices in what they build and what sources we have available. 
Re-adding the host devices is not a big deal. At the moment I don't think we are interested in acting as a host for a MTP or AUDIO device (well, we could connect a USB microphone, thinking of it). CDC devices are rare anyways.

m-chichikalov:
Have a question, Seems like macros under the F-buttons are not stored into the NON volatile memory, Am I right? If so, could you give me some clue where should I look to implement it?

db4ple:
They are stored. There is some code in ui_configuration.c to load and store. 
I never tested this AFAIR.

db4ple:
Before you go on, if not already done, please integrate the keyboard approach into the documentation in the Wiki: https://github.com/df8oe/UHSDR/wiki/Operating-Manual

m-chichikalov:
> Before you go on, if not already done, please integrate the keyboard approach into the documentation in the Wiki: https://github.com/df8oe/UHSDR/wiki/Operating-Manual

Sure...

db4ple:
Forgot to mention: https://github.com/df8oe/UHSDR/wiki/Digimodes#keyer-functionality is also worth looking at. A bit outdated now, thanks to your changes.

PR #1705, opened 2019-01-22
Removed unused but generated files from STM_HAL

db4ple:
- fixed location of host driver files in our tree
- fixed Eclipse config to know new include path for host driver headers
- aligned mcHF basesw usb_host.c with ovi40 / ovi40-h7 content


df8oe:
Now Makefile build is broken...

PR #1712, opened 2019-01-24
Refactored PSK and minor other changes

db4ple:
- PSK extensively commented
- PSK transmit now sends a preamble and postamble, works nice.
- PSK receive reworked to be easier to understand (but not working better, yet).
  Needs very (!) precise tune.
- Small changes to RTTY and PSK if used in combination with digibuffer
  buffered TX. In this case EOT is inserted automatically in the buffer
  and after transmission TX returns to RX properly.
- PSK code has less dependencies to rest of system. Also cleaned includes of
  RTTY.
- Fixed a small issue in digibuffer if keyer was used with PSK and full debug

### DIFF
--- a/mchf-eclipse/drivers/audio/psk.c
+++ b/mchf-eclipse/drivers/audio/psk.c
@@ -9,10 +9,63 @@
 
 // Common
 #include "psk.h"
-#include "ui_driver.h"
-#include "radio_management.h"
+#include <math.h>
 #include <stdlib.h>
+#include "softdds.h"
 #include "uhsdr_digi_buffer.h"
+#include "ui_driver.h" // only necessary because of UiDriver_TextMsgPutChar
+#include "radio_management.h" // only necessary because of RadioManagement_Request_TxOff
+
+
+// RX,TX common constants
+#define PSK_SAMPLE_RATE 12000 // TODO This should come from elsewhere, to be fixed
+
+// RX constants
+#define PSK_BND_FLT_LEN 5
+#define PSK_BUF_LEN (PSK_SAMPLE_RATE / PSK_OFFSET)
+// this must be an integer result without remainder and must be a multiple of 4
+
+#define PSK_SHIFT_DIFF (1.0 * PSK_OFFSET / PSK_SAMPLE_RATE) // phase change between two samples of PSK_OFFSET Hz
+
+// TX constants
+#define SAMPLE_MAX 32766 // max amplitude of generated samples
+
+
+typedef struct
+{
+    uint16_t rate;
+
+    uint16_t tx_idx;
+    uint8_t tx_char;
+    uint16_t tx_bits;
+    int16_t tx_wave_sign_next;
+    int16_t tx_wave_sign_current;
+    uint16_t tx_bit_phase;
+    uint32_t tx_bit_len;
+    int16_t tx_zeros;
+    int16_t tx_ones;
+    bool tx_win;
+
+    float32_t rx_phase;
+    float32_t rx_samples_in[PSK_BND_FLT_LEN];
+    float32_t rx_samples[PSK_BND_FLT_LEN];
+    int16_t rx_bnd_idx;
+    float32_t rx_cos_prod[PSK_BUF_LEN];
+    float32_t rx_sin_prod[PSK_BUF_LEN];
+    float32_t rx_scmix[PSK_BUF_LEN];
+    int32_t rx_idx;
+    int8_t rx_last_bit;
+    float32_t rx_err; // could be of interest for tuning
+    float32_t rx_last_symbol;
+    int16_t rx_symbol_len; // how many buffers fit into one bit
+    int16_t rx_symbol_idx;
+    // float32_t rx_symbol_buf[PSK_MAX_SYMBOL_BUF];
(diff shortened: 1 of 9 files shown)


mchf-eclipse/drivers/audio/psk.c
```diff
@@ -9,10 +9,63 @@
 
 // Common
 #include "psk.h"
-#include "ui_driver.h"
-#include "radio_management.h"
+#include <math.h>
 #include <stdlib.h>
+#include "softdds.h"
 #include "uhsdr_digi_buffer.h"
+#include "ui_driver.h" // only necessary because of UiDriver_TextMsgPutChar
```
m-chichikalov, line 16: Hi Danilo, I thought about it, we can remove it if switch UiDriver_TextMsgPutChar() to DigiModes_TxBufferPutChar( ... , BPSK );
And the same in RTTY. 
Theoretically should work....

mchf-eclipse/drivers/audio/psk.c
```diff
@@ -9,10 +9,63 @@
...
+#include "radio_management.h" // only necessary because of RadioManagement_Request_TxOff
```
m-chichikalov, line 17: I would like to handle it by adding the registering Callbacks interface in all digimodes:  TxRequest and RxRequest....  What do you think?

mchf-eclipse/drivers/audio/psk.c
db4ple: Just to say something about the current design and code:

The TxBuffer is a buffer serves as input for digital modes which require textual input.
This buffer is filled by one or more of the supported input methods (CW or Keyboard or Keyer). And read by a single consumer in an interrupt (by the active TX digi mode).
 
The UiDriver_TxtMsg buffer serves a input buffer for the console output for single producer of text in an interrupt, the output is in the main program part consumer.

The code for put and get  is safe to be used concurrently but not reentrant, i.e. while put and get may interrupt each other, it is not permitted to call put, while interrupt another put (same applies to get).  The code is non-blocking and require not explicit interrupt disabling or any other kind of synchronization or mutexes on critical sections.

For this reason these buffers shoud not be mixed. 

It is has not been the intention of these buffers to provide things like editable text (before it is sent) and to actually handle the display of both data on screen. That would be the purpose of a separate functionality on the UI side and should be decoupled from my point of view.

I agree that the UiDriver_TxtMsg buffer should not be part of the ui_driver.c and could be moved to the other buffer code serving the opposite direction or in a file nearby.

mchf-eclipse/drivers/audio/psk.c
db4ple: I don't understand what is your idea. Who is registering and who is calling the call backs?

A few comments:
The Tx Off/On Request functions just sets a flag, which tells the control logic to switch back to receive (or  transmit) at a convenient time. You can call this function anytime anywhere. 
I think, it is safe to say, we cannot decouple things much more than that.

If you mean to identify a common interface which all digi modes are supposed to implement, defining a structure of function pointers serving as interface to the modes (similar to what we do with the oscillators), well, that is possible and might be useful. 

I am fully into trying to identify common patterns, aligning signatures of functions serving a similar purpose, extracting shared code into separate functions in order to reduce code duplication and lower the maintenance effort where it makes sense, etc. All of this is a prerequisite for doing this kind of generalization. This was part of my motivation when I extracted the tx_processor code and refactored the code. However, looking at the code as it is, I see code duplication happening if we decouple things "too much" this way. 

I did the analysis which led to the two comments above mostly to understand why and how ui_driver is used here. Which should not be necessary. However, factoring out the buffer out of ui_driver is sufficient to solve the problem. A buffer like this should be exactly that, a buffer between different parts of the system which exchange data through the buffer asynchronously.

Since I don't have an answer for my initial question, I have not idea if your callbacks are something I am interested in or not. Generally I am not a too big of a fan for dynamically coupling everything since it makes navigating in the source code very difficult since you have to "execute" the dynamic registering in your head if you want to follow the path of execution just by looking at the source. 

That human execution is error-prone and takes more time to grasp than a direct call.  So at component boundaries it may make sense.

mchf-eclipse/drivers/audio/psk.c
m-chichikalov: Thank you for replay.

> Just to say something about the current design and code:
> 
> The TxBuffer is a buffer serves as input for digital modes which require textual input.
> This buffer is filled by one or more of the supported input methods (CW or Keyboard or Keyer). And read by a single consumer in an interrupt (by the active TX digi mode).
> 

Not only, Recording macros is also reading from it, not from interrupt as far as I know. 

> The UiDriver_TxtMsg buffer serves a input buffer for the console output for single producer of text in an interrupt, the output is in the main program part consumer.
> 

And here agoain, we have UI which produses ont from interrupt.
> The code for put and get is safe to be used concurrently but not reentrant, i.e. while put and get may interrupt each other, it is not permitted to call put, while interrupt another put (same applies to get). The code is non-blocking and require not explicit interrupt disabling or any other kind of synchronization or mutexes on critical sections.
> 
> For this reason these buffers shoud not be mixed.
> 

I see your main point, dont warry I keep it in mind.

Issue #1675, opened 2019-01-06
EEPROM corruption


sp9bsl:
Hello dear users,
we found an issue with EEPROM configuration memory. We found corruption of EEPROM signature in two devices: one mcHF and one OVI40 UI board. As a result of this behaviour there is "Wrong Signature" marking in MENU/System Info and lost of configuration data. 
If anyone experience this behaviour, please DO NOT REPLACE the memory with new one but report this fact here.

When the problem occurs, to help us resolve the problem please do the following:

- install a newer firmware which supports signature reset in the debug menu
- read the systems info message regarding eeprom signature, note down the shown value
- reset the EEPROM signatue
- dump the config data with the [UHSDR tool ](https://github.com/df8oe/UHSDR/wiki/Backup-and-Restore#backup--restore-on-external-computer-since-2919)
- upload the data here as ZIP

Thank you in advance!

df8oe:
I haven't seen this on any device yet. One scenaris where corruption can occur is "lost power during write".

It is possible to see if EEPROM is electrically damaged:
Do a "Rst Conf EEPROM" in Configuration Menu". This will write 0xff to all EEPROM cells - EEPROM now behaves as a brand new one. Now power down device using power button. Power on again, and after start, power down again using power button. If everything works as expected, your EEPROM must be recognized correctly and should be in use (green letters in System Information Menu).

sp9bsl:
Hi Andreas,
no, eeprom is not damaged, I found in my mcHF that signature (first byte of eeprom) is overwritten to 64 decimal, so writing it to 0xff and performing hardware probing completely resolved the issue. The second case (SP2GNB's OVI40 UI) is the same but I cannot determine the value of first byte, because we have tested the desoldered chip outside by write/read and forgot to dump it before... I know this issue is rare but occured in two different devices in a period of one week.

db4ple:
@sp9bsl: Could you please provide instructions how to preserve the right data for issue forensics right in your first entry so that the chance that someone with this problems finds out what to do?
Essentially, the user has to run or install a newer firmware which supports signature reset in the debug menu, read the systems info message regarding eeprom signature, note down the shown value , reset the EEPROM signatue and dump the config data with the UHSDR tool. And upload the data here as ZIP

m-chichikalov:
Ok, looks like I have such problem like this.
flash is 24CM02
In System info I see -> Wrong Signature/0B [not used]

"Rst Conf EEPROM" does not help.

[...]

PR #1713, opened 2019-01-25
Somehow related to #1675.

m-chichikalov:
Read comments under commits.

it's start - reading from eeprom to ram

it's copying back within changes applied
 

Do you see, it's writing back 0x00 and 0x00 to first two bytes.
I did not find what and where did not have time. almost night...
Added simple protection as a temp fix.

### DIFF
--- a/mchf-eclipse/misc/config_storage.c
+++ b/mchf-eclipse/misc/config_storage.c
@@ -142,14 +142,26 @@ uint16_t ConfigStorage_WriteVariable(uint16_t addr, uint16_t value)
     }
     else if(ts.configstore_in_use == CONFIGSTORE_IN_USE_RAMCACHE)
     {
-        uint8_t lowbyte;
-        uint8_t highbyte;
+        /**
+         * Was found that during saving settings,
+         * first two cells was corrupted.
+         * Just temp. protection...
+         */
+        if ( addr == 0 )
+        {
+            status = HAL_ERROR;
+        }
+        else
+        {
+            uint8_t lowbyte;
+            uint8_t highbyte;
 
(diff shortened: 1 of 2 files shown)


df8oe:
It must be a very, very rare situation. I do have 4 mcHFs / OVI40 here, I very often read and write configs to/from flash and to/from USB, do "reset serial EEPROM" and I *never ever* have seen any issue here...
So I will wait to merge anything here until the cause is identified 100%.

mmarzial:
I had the same problem 3 days ago, performing a reset with F1+F3+F5 keys sequence.
I had to remove the eeprom. it was a 24LC1025

db4ple:
@mmarzial:
F1+F3+F5 would not have helped in any case as it does not handle the wrong signature problem at all and is not intended to.
We introduced a new function for this in the Debug menu. However, even this one does not work yet if your signature is already destroyed. However, this was an programming error which we will fix soon.

mmarzial:
I wrote the comment because I got the eeprom corruption after F1+F3+F5 reset to default config

73 marco

db4ple:
Ok, thanks. Got it wrong. And it might indeed be the F1+F3+F5 could cause such a problem.
If you start with eeprom active and then press F1+F3+F5, we in the end wrote 0x0000 into the signature location.

With that in th eeprom, you get a wrong signature error and you were doomed (although we could have solved that without you desoldering the eeprom). Anyways, it is fixed now.

m-chichikalov:
It's also not checked boundary of ram_cache. it's easy overflow and corrupt some data in memory just by passing the wrong addr.  I added such this checking but not sure that I should push it.

PR #1740, opened 2019-02-08
Generalized ring buffer. 

m-chichikalov:
- general ring buffer for use across the code. re-entrant and multi-thread safe as long as we have two threads. Not portable probably, for H7 might be redone to use HW mutexes. Lock-free - mostly used pair of  LDREX and STREX instructions. Would be much easy just use __enable_irq() and __disable_irq to guard critical sections. Another way is to implement mutex but big question what should do Int. routine if it cannot obtain lock, deadlock.
@db4ple, I tried to imagine as much as I could scenarios in my head, seems good so far. Check it also, please. (It fails if more then two threads.) 
Tested only on F7.

- deprecated DigiModes_TxBufferHasData() to reduce the number of critical sections.

- in bpsk.c Eclipse automatically replaced tabs with spaces. I hope it's not a problem for anyone?

- (Eclipse config) Added generating Intel HEX as the build artifact. I need it for J-link.

- (Make) grep from CygWin does not work with --quit parameter. -q works fine.

- @db4ple Please, check where I put init digi_buffer. Maybe there is a more suitable place?


mchf-eclipse/misc/uhsdr_ring_buffer.c
```diff
@@ -0,0 +1,120 @@
...
+            retval = 1;
+        }
+        else
+        {
+            /** There is no room, write back old values */
+            next_head = (uint32_t)head_pair;
+        }
+    } while( __STREXW( next_head, ( volatile uint32_t* )&r_buf->head ));
+
+    /**
```
db4ple, line 96: What happens if another uhsdr_ring_buffer_put is executed right here after passing the while loop (i.e. in an interrupt of higher priority)?

m-chichikalov: it would use the next index as we already stored...
Let's assume we have to threads, main and an int.(int with higher priority)...
main execute  uhsdr_ring_buffer_put() ( tail = 20, head 30)
it read taill 20, compute next_tail 21, checking that we have room in buffer, stored 21 back and leave while loop... 

and here I probably need add memory barrier as it's the place where the changes should be visible for another observer.

Ok, we stored index. if it prevents by another thread here it would just use 31 index. 
Do not see an issue here.  If you see, could you share?

m-chichikalov: I see issue when we have two threads. P1 and P2. And P1 trying put, prevented by P2 here, P2 put into buffer, P2 continuet execution and later read from buffer. Here the issue, if in buffer just two entries( P1 is not finished puting ) it would read some data, but not what it should be there... 

Need think about it.

db4ple: Our low prio put enters with tail.idx/shdw = 20, head.idx/shdw = 30, so once we leave the guarded section we have our next_head at 31 and would have to store our data at 31, behind the last used slot. With __STREXW we store the next head in the head.idx   (which now reads 31, head.shdw == 30) 
Now we get interrupted with another put. Since we have left the guarded section, the higher prio put will have no trouble to execute. Meaning it will calculate to use slot 32 (since we have head.idx == 31 that this holds data) for storing, then stores the data here and finally advances head.shdw to 32.

And now we in the lower prio put get the control back. We execute the remainder of the code, retval is != 0, so we put our data in  r_buf->buf[ head_pair_ptr->index ] = data, which unfortunately is now pointing at 32. And we'll write our data there over the previously written data, but nothing has been written to slot 31. Which is bad.

I might be wrong, but it seems this algorithm does not fulfill is promise to be reentrant.
 
This hole subject is very, very difficult, that is why I try to stick to proven, well-known "text book" algorithms.

m-chichikalov: > Our low prio put enters with tail.idx/shdw = 20, head.idx/shdw = 30, so once we leave the guarded section we have our next_head at 31 and would have to store our data at 31, behind the last used slot. With __STREXW we store the next head in the head.idx (which now reads 31, head.shdw == 30)
> Now we get interrupted with another put. Since we have left the guarded section, the higher prio put will have no trouble to execute. Meaning it will calculate to use slot 32 (since we have head.idx == 31 that this holds data) for storing, then stores the data here and finally advances head.shdw to 32.
> 
> And now we in the lower prio put get the control back. We execute the remainder of the code, retval is != 0, so we put our data in r_buf->buf[ head_pair_ptr->index ] = data, which unfortunately is now pointing at 32. 

here head_pair_ptr->index still pointed at 31 (head_pair is located on the stuck of the prevented thread)...

If fallow books we should simply play with disabling and enabling irqs....

db4ple: You're right, I was confused a little by the creative use of casts to fake the pair data structure onto a single uin32_t. Nevermind, so this means we store the data at the "correct" location AFTER we stored data behind this location. Meaning we have for a certain period of time incorrect data in the buffer which cannot be detected by a get if it just hits at the right  moment, which could get the data from index 31 before it has been actually written.

m-chichikalov: the get would wait in loop until index and index_shodow would be equal. That's why, if put prevented by one put or one get it's fine, but if prevented by serial of put - get, get will read not stored et data.  - dead lock? Need think. What if an int. get() will wait main? Give me sec to think.

Now we have an unknown for users bug, that both main/an int routines put data into the buffer. So, seems that this approach solves this particular bug. Yes, it's not ideal, it's, as you said, hard to write 100% multi-thread, lock-free, re-entrant ring buffer... 

I would appreciate if you find time to read more carefully and analyze... Hope you find something that I do not see.

It's not rush.

BTW, do you know places where we use ring buffers? replace them by general one later?

m-chichikalov: Got an idea how to fix, will reopen later.

Issue #1750, opened 2019-03-02
Tune changes output level


K4LXY:
Your firmware version: 2.11.36
Your bootloader version: 5.0.4
UI Board: mcHF 0.6
RF Board: mcHF 0.6

After using TUNE, the power out level changes to that set for tune.
For example, I am using FULL power.  I push TUNE, which is set at 1 W.
When tuning is done, the output level is now set at 1 W instead of FULL.
I didn't notice at what firmware update this change happened.  
Not a big problem, however.
Tnx
Howard K4LXY

 * ...

K4LXY:
Thanks! Now power level returns to set value after tune.
73 Howard

K4LXY:
* UI Board: mcHF 0.6
 * RF Board: mcHF 0.6
Firmware 2.11.42
Bootloader 5.0.4

This is basically a followup on the issue I raised earlier, then closed.  In cw, if the power is set to, say, 5W, and the Tune level set to .5w, when I push Tune the power indicator says .5, but it actually puts out 5W.  This will only be seen if actual output monitor is monitored.

Again, certainly not a major problem.
73
Howard K4LXY

ase-2017:
Hi,
I can confirm the issue with 2.11.40. Will look into it.

K4LXY:
I'm measuring output with an external wattmeter (several, actually, and they all agree).
Howard

db4ple:
Hi Howard,
can you test 2.11.45 please, and report?
Thank you!

K4LXY:
Just tested.  When power is set at FULL and tune at .5, I get 1.5W out.  On the lower power settings (e.g. 5W, 2W), I get .5 out as it should.  I'd say it's working.   Thanks for working on this!
Howard

df8oe:
I do actually have access to 4 mcHFs. No one of this tunes with anything except "as TX power". Very strange. Additionally none of them shows anything on power meter - regardless of config "swap fwd/rev".

db4ple:
I'd say these may be very loosely related issues. 
Howard basically confirmed my measurements that 2.11.45 does the right thing.

[...]

PR #1761, opened 2019-03-15
Fix for Tune Power is not set correctly if not equal to current TX power (#1750)

db4ple:
we better tell the function which power we really want...

### DIFF
--- a/mchf-eclipse/drivers/ui/radio_management.c
+++ b/mchf-eclipse/drivers/ui/radio_management.c
@@ -230,11 +230,9 @@ float32_t RadioManagement_CalculatePowerFactorScale(float32_t powerMw)
  * @param band
  * @return true if the power factor value differs from previous
  */
-static bool RadioManagement_SetBandPowerFactor(band_mode_t band)
+static bool RadioManagement_SetBandPowerFactor(band_mode_t band, int32_t power)
 {
     float32_t   pf_bandvalue;    // used as a holder for percentage of power output scaling
-
-    int32_t power = ts.power;
 
     // FIXME: This code needs fixing, the hack for TX Outside should at least reduce power factor for lower bands
     if (band >= MAX_BANDS)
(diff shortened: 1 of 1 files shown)


df8oe:
After I merged this former funcionally "Tune Power" is destroyed again...

db4ple:
I have to check this again then.

df8oe:
Additionally: I do not see why my solution calls a function twice. It calls setting power at the beginning of transmission  (to set to desired level - setting of power is missing now) and then when going to RX it sets back to "standdard TX level".

db4ple:
I cannot see the issue reappear after line removal. 
It is working now as intended. I am building a 2.11.45 from unmodified source code.

The second call is "hidden" in the RadioManagement_SwitchTxRx() call.
You can check that with a ST Link, if you don't trust me and the code :-) . 
I was not saying it is obvious.

df8oe:
My build from source code of 2.11.45 tunes again with standard TX power and ignores TunePowerLevel while 2.11.44 works...

df8oe:
Maybe the value that is used in RadioManagement_SwitchTxRx() is the wrong one...

db4ple:
I debugged the function using ST-Link and it is using the correct one and the power meter also shows the right value. I'll test the official build this night and see if it behaves differently.

df8oe:
The meter is showing correct value in every case - but transmitting is always with "normal TX power" instead of "Tune Power"

db4ple:
How can the integrated watt meter (I think we are not talking about the xxW blue box), which measure the actual power output via ADC show the "correct"  value but the real output power is wrong.  The wattmeter measurements function / code is not related to the power.

I'll check this.

df8oe:
Integrated wattmeter does not show anything - it is always "0". I only can check at mcHF because all of my OVI40 boards are connected to experimental rf parts which do not have any PA stages.

df8oe:
Sorry - I have written misleading - I do mean TX power value display

df8oe:
Now for absolute clarifying this result for 2.11.45:

- Power value in display shows correct TX power in normal TX and in Tune function
- Power meter does show "0" (no bars) in all cases
- TunePower is always as normal TX power regardless of TunePowerLevel setting

tested with mcHF STM32F439 MCU and mcHF rf PCB.

db4ple:
My wattmeter (on OVI40 UI/ mcHF RF) shows correct value (in line with power value in blue box). It also doubles measured output if I for instance set tune to 2W instead of 1W. 

How do you measure the output power? With an external wattmeter?

Issue #1754, opened 2019-03-10
CW memory 2nd time


S52CQ:
Your firmware version: D2.11.36
Your bootloader version: 5.0.2
(Please note, we will generally not accept issue reports if your bootloader is not at least 5.0.1)

### Hardware 
 * UI Board: mcHF 6
 * RF Board: mcHF 6 

Hi,

CW memory kept working and not working so I have decided to find step-by-step setting that replicates the error.

1. Turn ON mcHF
2. Set VFO B. mode: USB, 7.074 MHz, set VFO A to 7.030, mode CW, long press M1, record a CW message (if not previously recorded), after that select VFO B
3. Turn OFF mcHF
4. After few seconds turn ON mcHF, it will have settings like in step 2.
5. Now select VFO A
6. Long press buttom M1 to enter CW memory
7. Play recorded CW msg: it will not play, it will just hang in TX doing nothing. To exit that, just press CW keyer.
8. Select different mode (LSB and back to CW: now all CW memories work well.

Best 73s de S52CQ, Jure

m-chichikalov:
Thank you, I will try to reproduce a bit later on this week.

S52CQ:
Hi Maksim,

any news on that issue?

73s de S52CQ, Jure

m-chichikalov:
HI, I was really busy these last two weeks on the field tests. Going spend some time on it during this week. I let you know. 73!

df8oe:
Does the fix work for you?

S52CQ:
I did not realize it was done. Will check it within few days and will 
let you know.

73s de S52CQ

S52CQ:
Yes, It works now!

Great job guys!

73s de S52CQ

m-chichikalov:
Hi @S52CQ, Seems like you reopened it, did you find something more?

S52CQ:
Hi guys,

today I was working S5 KVP contest. CW memories worked well. There are however two things that I'd like to check with you:
After (any) memory is keyed there is quite long delay when radio stays in TX mode - but it should as quickly as possible switch to RX because in contests things are happenig quickly. Is it just my radio or settings that I have wrongly set?
I have noticed that Memory 3 always stores a letter K o the ed, o matter what I record into it. Again, is it just my radio or am I doing womethig wrong? Memory 1 ad " are keying excacly what I have recorder into them.
73s de Jure, S52CQ

S52CQ:
[...]

PR #1766, opened 2019-03-27
Fix for - CW memory 2nd time #1754

m-chichikalov:
There is a kind of inconsistency in switching modes.
UiDriver_SetDemodMode() was not called during VFO changing.
Probably this was the cause of similar issues with other digi modes but not known yet.
It's the easiest way to fix it but as always is not the best one.

### DIFF
--- a/mchf-eclipse/drivers/ui/ui_driver.c
+++ b/mchf-eclipse/drivers/ui/ui_driver.c
@@ -1238,7 +1238,14 @@ void UiAction_ToggleVfoAB()
 
 	uint32_t old_dmod_mode = ts.dmod_mode;
 
+	/**
+	 * @warning - The two functions below call the same function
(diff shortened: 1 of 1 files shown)


db4ple:
What you do is to call the setting of the "new" mode unconditionally, in RadioManagement_ToggleVFOAB it is done conditionally. So the condition could not be correct (covers not enough cases) or ,as you wrote, we may have to look into the CW issue in more detail to find the root cause since I don't see why we should have to set the demod mode if we essentially just exchange the frequencies we're using but not the mode. It is similar to using the tuning knob in this case and should not require us to "reactivate" the mode.

m-chichikalov:
The changing VFO is not only about frequency, take a look into scenario described by [S52CQ](https://github.com/df8oe/UHSDR/issues/1754#issue-419176046)
You are right, it's done conditionally but not fully ( covers not all scenarios ).

PR #1804, opened 2019-08-24
CW bugfix for #1754

m-chichikalov:
Found couple bugs by adding unit tests for this cw state machine. More comming.
I'm prety sure the Iambic_B not working properly and Ultematic mode as well. 
The tests [here](https://github.com/m-chichikalov/uhsdr-unit-tests/blob/56ae6575f5a69bdc460488178e29ff550ae7cc34/tests/tes_cw_gen_host.cpp#L216).

### DIFF
--- a/mchf-eclipse/drivers/audio/cw/cw_gen.h
+++ b/mchf-eclipse/drivers/audio/cw/cw_gen.h
@@ -16,15 +16,15 @@
 
 #include "arm_math.h"
 
-// Exports
 void    CwGen_Init();
 
 void    CwGen_PrepareTx();
 void    CwGen_SetSpeed();
 
-bool    CwGen_TimersActive();
 bool    CwGen_Process( float32_t *i_buffer, float32_t *q_buffer, uint32_t size );
 
+// As showed tests calling them (CwGen_DahIRQ & CwGen_DitIRQ) is matter only for Ultimatic mode.
+// After change code to work in this mode w/o them, these extra dependencies with low-level layers could be removed.
 void    CwGen_DahIRQ();
 void    CwGen_DitIRQ();
 
--- a/mchf-eclipse/drivers/audio/cw/cw_gen.c
+++ b/mchf-eclipse/drivers/audio/cw/cw_gen.c
@@ -36,6 +36,7 @@
 // ----------------------------------------------------------------------------
 
 // Common
+#include "assert.h"
 #include "uhsdr_board.h"
 #include "radio_management.h"
 #include "softdds.h"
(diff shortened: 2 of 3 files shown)


df8oe:
Is line 243 of the tests intended as you want - or is it a typo? At line 259 there is following something very similar - but not identical. If this is intended: please can you rename line 243 so that it is not confusing for a typo?

mchf-eclipse/drivers/audio/cw/cw_gen.h
```diff
@@ -16,15 +16,14 @@
...
 void    CwGen_Init();
 
 void    CwGen_PrepareTx();
 void    CwGen_SetSpeed();
 
-bool    CwGen_TimersActive();
 bool    CwGen_Process( float32_t *i_buffer, float32_t *q_buffer, uint32_t size );
 
+// There is no real needs to call these functions. Going remove after adding more tests.
 void    CwGen_DahIRQ();
```
db4ple, line 28: What do you mean with this statement? These functions are called from the interrupt handler for the PTT/paddle GPIOs, see https://github.com/df8oe/UHSDR/blob/def4f73f1af6d081b2fb43cffb458f312f4fa6ed/mchf-eclipse/src/uhsdr_main.c#L62.

db4ple (changes requested):
Please follow the style guideline for code consistently: https://github.com/df8oe/UHSDR/blob/active-devel/CONTRIBUTING.md#coding-style-guide

Thank you!

db4ple:
BTW, interesting setup for testing. How can I setup this to build and run on my machine? 

EDIT: After looking into it, I can see that you are not running on target, only on the host for the moment. Got it.

m-chichikalov:
> Is line 243 of the tests intended as you want - or is it a typo? At line 259 there is following something very similar - but not identical. If this is intended: please can you rename line 243 so that it is not confusing for a typo?

Hi, thank you for review. Both of these test has similar "when" condition but they test different behavior. Everything fine there.

mchf-eclipse/drivers/audio/cw/cw_gen.h
m-chichikalov: > What do you mean with this statement? These functions are called from the interrupt handler for the PTT/paddle GPIOs, see
> 
> https://github.com/df8oe/UHSDR/blob/def4f73f1af6d081b2fb43cffb458f312f4fa6ed/mchf-eclipse/src/uhsdr_main.c#L62
> 
> .

I know, but answer why and for what exactly?

df8oe:
"Eclipse autoformat" is not what we want because it does not remove the newly added format issues but produces noise which makes it more difficult to identify code changes. Please revert the autoformat changes (they add some new format issues indeed) and remove the format issues following our  style guideline for code consistently manually at your added code.

Issue #1819, opened 2019-11-12
V2.11.71: missing 100kHz step


ghost:
Your firmware version: V2.11.71
Your bootloader version: 5.0.4

**Hardware**
SParrow with H7 CPU (400MHz version)

**Describe the issue:**
- missing 100kHz step

73!
Thomas

ghost:
solved - thx!

Thomas

PR #1821, opened 2019-11-12
 Fix for 100Khz Step not available #1819 

db4ple:
It will surely fix the 100Khz issue and should fix the 10khz issue as well but may not.


df8oe:
Nearly fixed now. Now only stuck at 10KHz when using STEP-. Rest works as expected.

PR #1824, opened 2019-11-16
Fix digit flickering (experimental)

db4ple:
The digit flickering for the main display should be gone now.
However, this is only a partial fix. Conceptually the flickering
still happens except for the main display using the 8bit font.
It is not noticeable for the small frequency display and no build currently
uses a non-8bit font for the main display.

It is still a kind of a mess until we support variable width fonts also for
the 1bit fonts or we have to draw the dots "by hand" between digits groups.


sp9bsl:
Hi Danilo,
the font data are generated automaticaly by software, but no problem now - I will modify the font editor software in the future to support this way.

db4ple:
@sp9bsl : Thanks! Good to know. I am not finally convinced that what I did is a good idea, so don't start with the modification now.

Issue #1828, opened 2019-12-15
Artefacts on screen when switching fro SPLIT back to normal operation


OK1FWG:
# UHSDR Firmware Defect Issue (Bug) Template 1.1

Please fill out the appropriate values.  Remove inapproriate/irrelevant values, but be prepare to provide more data. This template contains the most often asked questions.
We may adjust the template over time.

Please give as much information as necessary. At the same time, try to be concise. If you want to report something else than a defect, consider using a forum first. If you want report anything else but a defect remove the template if it does not make sense for your issue. Thank you!

## Issue Data goes here (please remove text above):

Your firmware version: 2.11.75
Your bootloader version: 5.0.4
(Please note, we will generally not accept issue reports if your bootloader is not at least 5.0.1)

### Hardware 
 * UI Board: mcHF 0.7
 * RF Board: mcHF 0.7

### Describe the issue:

Place tell us what does not work and how others can reproduce the issue (list related settings, steps to do, etc.):

### Your relevant settings
Hint: most of the information can be seen on a screenshot of the main display.

If you have display (also fill if audio problem) related problems:
 * Display resolution: 320x240 / 480x320
 * Display connected via: Parallel / SPI
 * Touchscreen: Yes
 * Magnify: 4x 
 * Spectrum: Dual
 

---> When switching from SPLIT operation back to normal, there are artefacts from the 'split' screen. It seem to me like new screen refresh is needed when switching to normal display to erase previous information from the buffer. It is directly behind the figures displaying the actual frequency. (If neede, I can provide a photos).

73, Willi, OK1FWG

df8oe:
Hi Willi,
thanks for the report. I can reproduce - but only using "modern font"

Bug has low priority.

vy 73
Andreas

db4ple:
@OK1FWG: Hi, did we fix this? Please verify and if yes, close the issue.

db4ple:
Closed due to inactivity

PR #1829, opened 2019-12-22
Fixed display artifacts when switch between split and normal mode display. (#1828)

db4ple:
Please test, I didn't test it on a real device yet.

### DIFF
--- a/mchf-eclipse/drivers/ui/ui_driver.c
+++ b/mchf-eclipse/drivers/ui/ui_driver.c
@@ -1590,7 +1590,10 @@ void UiDriver_DisplayDemodMode()
 	UiDriver_DisplayModulationType();
 }
 
-
+/**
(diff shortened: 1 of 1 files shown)


df8oe:
There are still artifacts. But they are different to the ones before:
Old font:
- there are no artifacts using former firmware
- using this fw there are artifacts in the "split screen" right to the ">" symbols in RX and TX
Modern font:
- there are artifacts around the two decimal points in single frequency display and no artifacts in "split screen"
- using this fw there are only artifacts at the left decimal point in single screen but as mentioned above additionally artifacts in "split screen".

df8oe:
Isn't it best solution to write a black rectangle at the complete frequencydisplay position when changing from split to standard and reverse? The time which is needed is very low and it will work whatever font in the future may be added...

db4ple:
That is exactly what we are trying to do (drawing a black rectangle covering the full area) with this commit. 
However, the rectangle size was calculated in the wrong way. Hence no effect (or compared to the previous version a worse effect).

PR #1834, opened 2020-01-31
hex2dfu changes for python3

csylvain:
my system is prepared for building fw and has only Python 3.
these changes update the tool used in the UHSDR fw build recipe.

### DIFF
--- a/mchf-eclipse/Makefile
+++ b/mchf-eclipse/Makefile
@@ -215,7 +215,7 @@ CC      = $(VPRE)$(PREFIX)arm-none-eabi-gcc
 CXX     = $(VPRE)$(PREFIX)arm-none-eabi-g++
 OC      = $(VPRE)$(PREFIX)arm-none-eabi-objcopy
 SZ      = $(VPRE)$(PREFIX)arm-none-eabi-size
-HEX2DFU = $(VPRE)python $(ROOTLOC)/support/hex2dfu/hex2dfu.py
+HEX2DFU = $(VPRE)python3 $(ROOTLOC)/support/hex2dfu/hex2dfu.py
 
 POSTCOMPILE = @mv -f $*.Td $*.d && touch $@
 DEPFLAGS = -MT $@ -MMD -MP -MF $*.Td
--- a/mchf-eclipse/support/hex2dfu/hex2dfu.py
+++ b/mchf-eclipse/support/hex2dfu/hex2dfu.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
 #
 # Intel HEX to ST microelectronics DfuSe file converter
 # Copyright (C)2015 Thomas Kindler <mail@t-kindler.de>
@@ -29,7 +29,7 @@ def load_hex():
     :return:    intelhex object
     """
     if args.verbose:
-        print "Loading %s..." % args.source
+        print ("Loading %s..." % args.source)
 
     try:
         ih = IntelHex()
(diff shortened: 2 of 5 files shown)


df8oe:
It is a very good idea to implement python3! But because there is many, many stuff that has not migrated to python3, python on most systems targets python2 (so on my bleeding edge "Siduction" that bases on Debian experimental). So I think it would be wise not to point to
/usr/bin/env python
but to
/usr/bin/env python3...

csylvain:
suit yourself (as in: suits me fine either way). i've had too many headaches keeping both pythons around. whenever i can, i now leave python2 out of the picture. helps to move the migration along, yes?

csylvain:
made the suggested change. found something else needing a change, too.

df8oe:
You have to look again. .dfu-files are much too long - there is definitely something going wrong... Normally difference between bin and dfu is max. 500 bytes. The now created dfus are more than 11KB large (bootloader build)...

csylvain:
okay.  will have to be much more in depth than fixing syntax problems.
with 2.11.78 and python3, i have
```
$ du -sh fw-*
420K    fw-mchf.bin
572K    fw-mchf.dfu
2.7M    fw-mchf.elf
768K    fw-mchf.map

**Comment by csylvain** [comment]
with 2.11.78 and python2, i have
```
$ du -sh fw-*
420K    fw-mchf-python2.bin
420K    fw-mchf-python2.dfu
2.7M    fw-mchf-python2.elf
1.3M    fw-mchf-python2.hex
768K    fw-mchf-python2.map

db4ple:
Hi, @csylvain: 
Thanks for doing the migration to python3 
 
The issue in your changes is here:
https://github.com/df8oe/UHSDR/blob/1c64ad45e7a2b81480b24f7051cab0338ee5fd31/mchf-eclipse/support/hex2dfu/hex2dfu.py#L73
You cannot have python translate a binary string using the utf-8 encoding into a byte array. It will expand some of the binary sequences into the related utf-8 encodings which uses more bytes, hence the longer resulting binary.

A question for you: Did you test the resulting dfu files produced by your code?

Good news is, I think to get the same result on python2 and python3 a simple change will suffice:

```python
    image_data = ih.tobinarray()

    data = struct.pack(
        "<II",
        ih.minaddr(),       # dwElementAddress
        len(image_data)     # dwElementSize
    ) + bytes(image_data)          # Data
````
 
This, fortunately makes the code run on python2 and python3 with the same result.

csylvain:
sorry, the utf-8 got added because your suggestion (which i tried at first using the ascii encoding) only produced an error indicating it was unable to create the bytes without an encoding.
now that i have created and preserved a python2 baseline, i will investigate further.

db4ple:
@csylvain: I, db4ple, did not suggest anything. But the important thing is to switch to use `ih.tobinarray()` instead of `ih.tobinstr()`, which allows to use bytes() without re-encoding.  
BTW, the makefile calls the script with /usr/bin/python and bascially ignores the #! setting.
But with the script running on both python2 and python3 there is no problem with that.

csylvain:
i thought i recognized the code snippet and also that you were pointing only at the bytes() call.
in that stanza, then, there were two changes to make and i made only one. okay, then.
shall i work up a new pull request against 2.11.78 ?
also, as someone may want to call hex2dfu on it's own, i will leave the python3 on the hashbang line.

db4ple:
@csylvain: No problem, my fault. But again, good job. You should have tested the resulting dfu, though.
@df8oe: Good job spotting this.
BTW, despite my previous message, the code does not run on python2 and python3 in identical form.
Pyhon2 -> ih.tobinstr plus bytes works, ih.tobinarray produces very large output
Python3 -> ih.tobinarray plus bytes works, ih.tobinstr requires encoding in bytes and produced wrong output.

PR #1835, opened 2020-02-02
mistake corrected in hex2dfu - do not preserve use of tobinstr() pyth…

csylvain:
…on2-ism - use tobinarray()
with this done, 2.11.78 and python3 produces
```
$ du -sh fw-*
420K    fw-mchf.bin
420K    fw-mchf.dfu
2.7M    fw-mchf.elf
832K    fw-mchf.map
```
the difference between the dfu file and the bin is 309 bytes.

### DIFF
--- a/mchf-eclipse/support/hex2dfu/hex2dfu.py
+++ b/mchf-eclipse/support/hex2dfu/hex2dfu.py
@@ -66,6 +66,10 @@ def save_dfu(ih):
     #
     image_data = ih.tobinarray()
 
+    if len(image_data) != len(bytes(image_data)):
+        # catch and correct python2.7 issue using this hack
+        image_data = ih.tobinstr()
+
     data = struct.pack(
         "<II",
         ih.minaddr(),       # dwElementAddress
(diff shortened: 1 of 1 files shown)


mchf-eclipse/support/hex2dfu/hex2dfu.py
```diff
@@ -64,13 +64,13 @@ def save_dfu(ih):
 
     # Image element
     #
-    image_data = ih.tobinstr()
+    image_data = ih.tobinarray()
```
db4ple, line 67: Please add this, it does produce identical output on both python2 and python3:
```
    if len(image_data) != len(bytes(image_data)):
        #python 2 hack
        image_data = ih.tobinstr()
```

csylvain: sure thing. einen Moment bitte  :)

df8oe:
I have not tested if .dfu is working yet  - will do that later. But now it is looking fine! Thank you for migrating to python3.

csylvain:
you are welcome. in any case, now both the python2 baseline and the files made with the hex2dfu port to python3 are exactly the same number of bytes: 429040 (.bin) and 429349 (.dfu), respectively.

csylvain:
built a fw today with a new feature added.  mcHF booted up fine and so far so good on testing. pull request in the works

PR #1836, opened 2020-02-04
adds MIC_TYPE menu item; uses existing unreferenced MIC_BOOST

csylvain:
"adds MIC_TYPE menu item and uses existing (unreferenced/unused) MIC_BOOST EEPROM location for switchable gain value"

EEPROM_MIC_BOOST was not created; it already exists. this commit adds code to use it

in struct TransceiverState, a new uInt8_t variable tx_mic_boost was created
 
"Mic Type" menu selection of "Dynamic" or "Electret" is placed in the Standard Menu just ahead of the "Mic Input Gain" menu item

selecting "Dynamic" sets the tx_mic_boost value to 15 (15 dB gain)
selecting "Electret" sets the tx_mic_boost value to 0 (no additional gain)

when tx_mic_boost > 0 an additional 31.6 is added 31.6 to gain_calc in a way which does not disturb the existing mic audio gain and incoming stream processing

fw "X2.11.78" has been built and basic testing done

(i may want to tweak the Dynamic Mic boost value a bit. here a Mic Gain of 60 peaks 'just right' but with a bit of tweaking further the default setting of 35 could be made equivalently 'just right')


mchf-eclipse/drivers/ui/menu/ui_menu.c
```diff
@@ -626,7 +626,7 @@ const char* UiMenu_GetSystemInfo(uint32_t* m_clr_ptr, int info_item)
       		snprintf(out,32, "D%s", UHSDR_VERSION+4);
 		  #endif
 		#else
-    	  snprintf(out,32, "%s", UHSDR_VERSION+4);
+			snprintf(out,32, "X%s", UHSDR_VERSION+4); // eXperimental fw build
     	#endif
```
df8oe: Why do you think every build which is not made by me is "experimental"? I do not think so... They are not official builds so trailing "D" misses. But no other trail is needed.... Pse remove this.

csylvain: seemed like a good idea at the time. reverted. gain tweaked. more gain was not better especially when CMP is enabled

csylvain:
thanks!

df8oe:
If you want I will add your callsign to the "Hall of Fame" next release... I hope you will stay at UHSDR and contribute more in the future ;)

csylvain:
that would be nice, thanks!   i'm working on another contribution right now.

  - \0111 -

df8oe:
Will be seen in next binary release. Thank you for your contribution! BTW: There is another python usage - maybe it is interesting for you to port it to py3, too?

csylvain:
hello!  i've submitted a new issue to let you know i've updated the wiki.
spelling corrections, and incorporating the images in a new section from
EA5HZZ's gitbook in a way that makes the new files self-contained within
the wiki's repo. i can eventually comb through the wiki and do the same
with the other image file references to the UHSDR code repo.

i had to consult with github people to figure out the correct way to use
the MediaWiki syntax of Markdown to do it.

on my "To Do" list:

1. when saving firmware from the mcHF to pendrive, it does not supply a
valid date value for the firmware file. this causes an error message when
listing the files on the pendrive ("dir" command). i intend to hardcode a
reasonable file date like January 1, 2019 for now as a fix and submit a
pull request.

2. i believe i have found the reason why the digital audio (DIG EXT) level
is not maintained at a constant level when a constant level audio is
provided. the SSB audio processing does not distinguish between microphone
input and usb input and does the audio AGC for both. i intend to add a
check and skip the audio AGC when the mode is SSB and the audio input is
DIG EXT.

3. what is the other python tool needing a port to python3 ?

ciao!

Issue #1843, opened 2020-04-14
menu_handbook build files need update to python3


df8oe:
My workstation where UHSDR builds are done now uses python3 as default. Handbook builds now stop with

Traceback (most recent call last):
  File "./ui_menu_structure_c2py.py", line 37, in <module>
    BUILD_ID = MAJ + "." + MIN + "." + REL
TypeError: can't concat str to bytes
Traceback (most recent call last):
  File "./ui_menu_structure_graph.py", line 11, in <module>
    from ui_menu_structure_c2py import MENU_DESCRIPTOR, BUILD_ID
  File "/home/andreas/git/uhsdr/mchf-eclipse/support/ui/menu/ui_menu_structure_c2py.py", line 37, in <module>
    BUILD_ID = MAJ + "." + MIN + "." + REL
TypeError: can't concat str to bytes
Traceback (most recent call last):
  File "./ui_menu_structure_mdtable.py", line 12, in <module>
    from ui_menu_structure_c2py import MENU_DESCRIPTOR, BUILD_ID
  File "/home/andreas/git/uhsdr/mchf-eclipse/support/ui/menu/ui_menu_structure_c2py.py", line 37, in <module>
    BUILD_ID = MAJ + "." + MIN + "." + REL
TypeError: can't concat str to bytes

db4ple:
Hi,
this is a common update issue as python3 changed a couple of this in the str vs. bytes handling.

PR #1848, opened 2020-04-21
python2 and python3 compatible handbook generation (fix for #1843)

db4ple:
Any python should work now (tested with python2 and python3)
Default is just to call python, but with
`PYTHON=python2 make handbook-ui-menu-clean handbook-ui-menu`
or
`PYTHON=python3 make handbook-ui-menu-clean handbook-ui-menu`
you can make it call whatever python you want.


df8oe:
The exec bit for mk-menu-handbook was lost. I already fixed that.

PR #1854, opened 2020-05-16
Dockerfile URL missing release candidate info.

innattar:
I was unable to complete all the steps in CONTRIBUTING.md for setup on gitpod.io.  The dockerfile build step was failing on wget.  After some digging I realized the release candidate info was missing from the URL.


db4ple:
Strange, was not necessary when I set this up.  Thanks for testing. 
They changed the release URL after release and it is not fitting the usual scheme. Let's merge this.

PR #1855, opened 2020-05-16
VSWR foldback protection only when min power available

innattar:
While testing SSB transmit on the MCHF I kept having VSWR protection kick in (blinking TX).  This would not happen during tune or if I used a 100% duty cycle mode like FM.  After troubleshooting I realized this only happened when the transmit power level dropped.  The UI code set precedent for not computing SWR (for display purposes) if a minimum forward power was not met.  I took this concept and applied it to the VSWR foldback protection code.   I tested this on an MCHF under the same conditions as before and had no issue with VSWR protection kicking in.

### DIFF
--- a/mchf-eclipse/drivers/ui/radio_management.c
+++ b/mchf-eclipse/drivers/ui/radio_management.c
@@ -38,6 +38,7 @@
 
 #include "ui_configuration.h"
 #include "ui_menu.h"  // for CONFIG_160M_FULL_POWER_ADJUST and Co.
+#include "ui_driver.h" // for SWR_MIN_CALC_POWER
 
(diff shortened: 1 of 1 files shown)


innattar:
Sorry this is my first time collaborating on Github.  This PR is meant to be stacked on top of https://github.com/df8oe/UHSDR/pull/1854.  Also I should note that I chose to split this up since the two changes were not related and I thought it would be best to have two PRs.

Since that isn't in the main repo yet both commits are showing up in the commit list for this PR.  I apologize for any confusion.

innattar:
Thanks for merging #1854.  I have rebase and pushed up this branch so it only includes the one change.

PR #1861, opened 2020-05-19
Fix compiler warnings regarding abs function and float/int arguments.

innattar:
Just for fun I tried building with `gcc-arm-none-eabi-9-2019-q4-major`.  There were compile warnings in the FreeDV code that were surfaced by `-Wabsolute-value`.  As a fix, if the variable being assigned to was a float I used the `fabsf` function otherwise the float math was cast to `int` before passing to the function.  For the latter two lines of code this is likely a no-op (but at least it tells the reader the behavior is expected).  For the first line of code this may change the output since float->int->float conversion no longer occurs (this can have an impact on decimal precision of the resulting float).  Also on line 299 the abs function was called twice.  Since abs() is idempotent calling it once is sufficent.

I tested this by performing a FreeDV 1600 QSO between the MCHF and a Yaesu991A+Raspberry Pi running the latest FreeDV.  No discernable difference detected between audio and no issues decoding Text Frame.

Note, while the issue was discovered by building with `gcc-arm-none-eabi-9-2019-q4-major` all tests were run with firmware built using `gcc-arm-none-eabi-8-2019-q3-update`.

### DIFF
--- a/mchf-eclipse/drivers/freedv/c2wideband.c
+++ b/mchf-eclipse/drivers/freedv/c2wideband.c
@@ -257,7 +257,7 @@ void correct_rate_K_vec(MODEL *model, float rate_K_vec[], float rate_K_sample_fr
             float cka = K;
             int k;
             for (k = 0; k < K; k++) {
-                float a = abs(rate_K_sample_freqs_kHz[k] - nasty_error_freq);
+                float a = fabsf(rate_K_sample_freqs_kHz[k] - nasty_error_freq);
 
                 if (closest_k == -1 || a < cka) {
                     closest_k = k;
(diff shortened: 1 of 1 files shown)


db4ple:
Hi @innattar, thanks for the fix, but this is the wrong place for it. Please submit to the codec2 main repository: https://github.com/drowe67/codec2

We are pulling this code straight in from there from time to time, so your fix would be lost as soon as we do this...

The codec2 guys are very friendly and will appreciate your contribution, even if it is small...

innattar:
@db4ple I took a quick look at `codec2` and it appears that this code has changed significantly to the point where these lines no longer exist.  Thank you considering this PR.

PR #1875, opened 2020-09-17
Fix bootloader build issue for Eclipse

db4ple:
added missing include path to ALL languages in all bootloader configurations


df8oe:
I do not understand why this gets merge conflicts... You have chenaged one line and added one line on serveral places. Why can GitHub not detect that? I can solve the conflict by adding the lines manually. But I do not know if this is a good idea or if you can solve the conflict in your pull request...

db4ple:
There is one detected conflicting change in/around line 1328:

<listOptionValue builtIn="false" value="/mchf-eclipse/drivers/usb/host/class/HID/Inc"/>

I don't understand it either. That's why I had to submit the CR (to be able to see the conflict).

I will fix this.

db4ple:
Now the PR has no conflicts anymore.

PR #1879, opened 2020-11-10
Python3 change for uhsdr.py

w1mt:
Had to change uhsdr.py so that the bytearray cast would work. I think this is a necessary Python3 change. I have tested this for backup and restore and it seems to work for me. Thank you for all your work!

If something about this falls outside the normal workflow feel free to remedy as you see fit.

-mike/w1mt


db4ple:
Approved. Thanks, works for python2 and python3 as well.

Issue #1895, opened 2021-01-31
GCC 10 at least with Eclipse MCU does not link firmware


db4ple:
## Issue Data goes here (please remove text above):

Your firmware version: current active-devel 2.11.95

### Hardware 
all
### Describe the issue:

Firmware build fails during link process with error regarding multiple instances of functions / c data structures.
Firmware compiles just fine if gcc 8 or 9 is being used in the same environment.

Current workaround is to use  gcc 8 or gcc 9 as compiler.
### Your relevant settings
Build environment Eclipse 2020-12 Embedded CDT plus xpacks GCC 10.2 on Ubuntu

df8oe:
I can confirm this using gcc10 with Makefile

I am using gcc-arm-none-eabi-8-2019-q3-update because of this version produces the smallest binaries and no other version of gcc delivers binaries which reach this size in any way. I am not talking of 1...2 % - I am talking of 30...50 %. I do not know if there is a way to reduce binary size if using higher gcc - but the size of the binary is very important for us...

PR #1896, opened 2021-02-12
Fix for  #1895 GCC 10 does not link firmware

db4ple:
Fixed a couple of programming issues which have been masked
/ not cared about in gcc 9 and earlier...

Should work as good as before


df8oe:
I can confirm gcc8 is building and linking smoothly as before.

PR #1903, opened 2021-02-25
preparations for Fast Convolution Filtering #1

DD4WH:
1st preps, not functional

### DIFF
--- a/mchf-eclipse/drivers/audio/audio_convolution.c
+++ b/mchf-eclipse/drivers/audio/audio_convolution.c
@@ -20,6 +20,31 @@
 // use FreeDV buffers, if we switch on digital voice mode FreeDV
 // user code
 
+/***************************************************************************************************
+ *
+ *  Fast Convolution Filtering
+ *  2nd attempt, DD4WH, 2021-02-25
+ *
+ *  should run with 24ksps sample rate in order to be able to allow large bandwidths up to 11kHz
+ *  no partitioned convolution necessary, just plain simple Fast FIR convolution is sufficient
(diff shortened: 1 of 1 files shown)


mchf-eclipse/drivers/audio/audio_convolution.c
```diff
@@ -921,7 +904,7 @@ void AudioDriver_RxProcessorConvolution(AudioSample_t * const src, AudioSample_t
 
     // interpolation
 
-// TODO: at this point we have 128 real audio samples filtered and AGC�ed in the variable 	 (for mono modes)
+// TODO: at this point we have 128 real audio samples filtered and AGC�ed in the variable 	 (for mono modes)
```
df8oe, line 907: Is it possible that you use a character set which is not UTF-8? Here it is only a simple character - may grow intensively at other files. Please check.

db4ple: It can be set per file. but best set this for the whole workspace: 

https://stackoverflow.com/questions/9180981/how-to-support-utf-8-encoding-in-eclipse

https://ralph.blog.imixs.com/2019/05/07/eclipse-utf-8-encoding/

DD4WH: strange, this question mark character has been in the code for at least 3 years and I did not change it.
My settings were and are "UTF-8"
Which action should I take now?

df8oe: Strange indeed! I will merge it - it is not problematic at all. Does Eclipse support charecter set checking? We should check that some time - not important now. I only wanted to put your focus on your encoding settings if they are right.

PR #1910, opened 2021-04-09
Fix macro to ensure a comma at the end

howard0su:


db4ple (changes requested):
Hi, if we touch this, we should it right: Use of an static inline function.
```
static inline void GPIO_SetBits(GPIO_TypeDef *PORT, uint32_t PINS) { ... }

What do you think of this?

howard0su:
Yes, agree if this is the way we want to go with. let me do it.

db4ple (changes requested):
This code will not compile on gcc10  as the compiler now requires `inline static` where `inline` could be used previously if the same inline function was included in multiple compilation units.

howard0su:
sure. i will change it. but there are other inline function in this file is not static inline as well.

which version of gcc10 will require this? My local gcc works fine.
howardsu@macpro mchf-eclipse % arm-none-eabi-gcc --version
arm-none-eabi-gcc (GNU Arm Embedded Toolchain 10-2020-q4-major) 10.2.1 20201103 (release)

db4ple:
Hi, it is the gcc 10 2020 q4, as now used in Travis.  See #1896 (profiling.h) . The problem only appears if the compiler for whatever reason decides not to inline the code and this happens in multiple files. For the one-liner it probably never happens but it is easily avoided that it ever happens by adding static where appropriate. In fact `inline extern` is also an interesting option, but I think we don't need this here. 

Anyways, appreciate that you followed the suggestions. Obviously it would have been easier to accept and then directly change the code but I thought, you would want to make a nice submission and would like to have your name in the "blame" history shown as you identified the improvement area.

howard0su:
no problem. love to listen your suggestions.

PR #1911, opened 2021-04-09
Create Github action script

howard0su:
This script will run when there is PR or Push to devel branch to verify all 3 bootloader builds and 7 firmware builds.

I use a set of makefile to capture the difference of boards. And the build will run parallel.


howard0su:
The action script almost do everything as trivas-ci. I didn't add doc build which I don't think we need it for every PR.

The biggest benefit is build time reducing from 15mins to 3mins.

howard0su:
from the other thread, maybe we want to have different GCC version to test the build.

db4ple:
I think, it is helpful to be able to choose which compiler is being used. Currently we use two of them: The newest (right now GCC10) for development (I switched compiler to gcc10 in travis to reflect this) and GCC8 for compiling release binaries as this for yet unknown reasons creates much smaller binaries compared to earlier or later gcc versions with exactly the same settings. The size is most important for users with the old STM32F4 with 512k flash only, as on these machines a larger binary will not fit without reducing the included functionality. 

@howard0su: Can you point me to a link where I can see a log of your action being run on a Github repo?

howard0su:
https://github.com/howard0su/UHSDR/actions/runs/732308684

I am using the built-in gcc compiler of ubuntu

df8oe:
As Danilo already stated: we should be able to use different compiler versions for the different build processes. And as I already stated: we already have a working automated workflow to build the binaries without dependence to any external service and load them up to a webspace which cannot be cancelled or limited from external...

